PR #51, opened 2018-11-06
#49 docker instead solr wrapper

cdmo:
1. changes namespace to solr from psulib_traject:solr
2. adds 3 tasks, up, clean and stop
3. updates documentation accordingly

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,7 @@ GEM
       public_suffix (>= 2.0.2, < 4.0)
     builder (3.2.3)
     byebug (10.0.2)
+    columnize (0.9.0)
     concurrent-ruby (1.0.5-java)
     domain_name (0.5.20180417)
       unf (>= 0.0.5, < 1.0.0)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -88,9 +74,8 @@ DEPENDENCIES
   marc
   rake
   rsolr
-  solr_wrapper!
   traject (= 3.0.0.alpha.2)
```
banukutlu: traject 3.0 did not work?

banukutlu:
```bundle exec rake solr:up
felix
rake aborted!
Don't know how to build task 'docker:ps' (see --tasks)
```

need this task: 

```
  task :ps do
    print `docker ps`
  end
```

banukutlu:
I added `require 'faraday'` for solr:conf to work and removed the obsolete solr conf files from the list of files to be downloaded

Gemfile.lock
banukutlu: with 1f463ce18 this is solved

cdmo: hmm thought i did a bundle install... maybe i need to specifically do bundle update?

cdmo: so that is outdated, i think i caught this mistake after-the-fact

cdmo:
I am getting that too. Not sure why it's returning a 400 *and* successfully clearing the index. Maybe docker something something.

PR #57, opened 2018-11-26
#14 index and display titles

### LINKED ISSUE #14: Index and display titles
In GitLab by @rkt6 on Aug 16, 2018, 11:35

https://github.com/psu-libraries/psulib_blacklight/wiki/Title-Fields-Indexing-and-Display

To-do:

* [x] Create title statement for display. `245abcfgknps` including vernacular. These should be used for the main title on a record page and the linked title in a search result. Vernacular must be displayed in search result snippets
* [x] Create a uniform title for display. `130adfklmnoprs:240adfklmnoprs:730ail` including vernacular. Uniform titles are a common way of referring to a work which can be called many things.
* [x] Create additional titles display. `210ab:246iabfgnp:247abcdefgnp:740anp` including vernacular. These include subtitles former titles, and uncontrolled versions of the title.
* [x] slice ` --` off this in cases from 505 for related titles `title_related_tsim`? Trim punctuation doesn't handle it

To-review (already in Traject and we should decide how we use and whether we remove): 

* [x] title_t 245a - first part of title. Can/should be used for exact matches and for general title index - to be handled in https://github.com/psu-libraries/psulib_blacklight/issues/180
* [x] title_display / title_vern_display - 245a. May want to deprecate.
* [x] subtitle_t - 245b. Subtitle. Unsure whether it will be useful. May need to deprecate.
* [x] subtitle_display / subtitle_vern_display - 245b, came with traject. Deprecate.

Required by https://github.com/psu-libraries/psulib_blacklight/issues/180

banukutlu:
Closes #14


psulib_config.rb
```diff
@@ -92,17 +97,70 @@
...
+  797lktmnoprs
+  798lktmnoprs
+  799alktmnoprs
+].join(':'), trim_punctuation: true) do |record, accumulator|
+  accumulator.each { |value| value.chomp!(' --') } unless record.fields('505').empty?
+end
 
-to_field 'title_series_tsim', extract_marc('440anpv:490av')
+## Title Display Fields
+to_field 'title_latin_display_ssm', extract_marc('245abcfgknps', alternate_script: false, trim_punctuation: true)
```
cdmo: Maybe this should be just `title_latin`?

psulib_config.rb
```diff
@@ -92,17 +97,70 @@
...
+to_field 'title_vern', extract_marc('245abcfgknps', alternate_script: :only, trim_punctuation: true)
+# use vern title as title_display_ssm if exists
+# otherwise use latin character title as title_display_ssm
+each_record do |_record, context|
```
cdmo: If my understanding is correct, `each_record` is only called once and can be used to modify any field. Right? With that maybe we should push it down to the bottom of the config file? What do you think?

psulib_config.rb
banukutlu: this field is displayed in the item level main content when the vernacular is the main title so it is part of the index 🤔

banukutlu: I like your approach, let me test it on my local.

cdmo: I'm not sure if one way is more efficient, I would guess your way is slightly faster, fwiw. Again, both ways are totally fine IMO.

psulib_config.rb
banukutlu: since it adda a new title field, kinda nice to have all title related fields together as well 🤔

cdmo: 👌

Issue #18, opened 2018-10-03
Mapping publication fields


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Publication-Info-Fields-Index-and-Display

banukutlu:
In GitLab by @rkt6 on Sep 4, 2018, 15:51

assigned to @rkt6

banukutlu:
In GitLab by @rkt6 on Sep 5, 2018, 10:41

changed the description

banukutlu:
In GitLab by @rkt6 on Sep 5, 2018, 10:42

changed the description

banukutlu:
In GitLab by @rkt6 on Sep 5, 2018, 10:42

changed the description

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:08

created branch [`18-index-and-display-publication-fields`](https://git.psu.edu/i-tech/psulib_traject/compare/master...18-index-and-display-publication-fields)

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:18

Fields added:
- production_display
- distribution_display
- manufacture_display
- copyright_display
- edition_statement
- pub_location_t
- publisher_t
- manufacturer_t

(note, contents of `published_display` and `published_vern_display` were changed)

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:19

marked the task **Publication date** as completed

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:19

marked the task **Copyright date `264|*4|c` (trim punctuation)** as completed

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:19

marked the task **Manufacturer - `260f:264|*3|b` (trim punctuation)** as completed

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:19

marked the task **Publisher - `260b:264|*1|b` (trim punctuation)** as completed

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:19

marked the task **Publisher location - 260a:264a (trim punctuation)** as completed

banukutlu:
In GitLab by @rkt6 on Sep 7, 2018, 12:20

marked the task **Publication/edition display** as completed

banukutlu:
[...]

PR #85, opened 2019-01-24
mapping publication fields #18

banukutlu:
Connected to #18


banukutlu:
> Was the Gem update incidental? Or required for this task?

not required for this task. I somehow could not get rubocop to work but after the update it worked.

PR #89, opened 2019-01-31
#31 publication date

### LINKED ISSUE #31: mapping publication year facet
In GitLab by @rkt6 on Sep 7, 2018, 12:20

Does the marc_publication_date function produce what we actually need for our publication date information? Review what the process does and how well it fits for our use.

Related to #18

banukutlu:
Closes #31 

`process_publication_date` is based on Traject's `marc_publication_date` but we resort to `264|*1|c` then to `260|c` if no acceptable date found in `008` field.

`process_publication_date` can be refactored further - but just to be able to wrap up the publication fields, I am adding the PR right now but curious to hear about your feedback @cdmo.


.gitignore
```diff
@@ -12,3 +12,7 @@
 
 # testing output
 coverage
+
+# ignore sample data except sample_psucat.mrc
```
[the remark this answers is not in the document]
cdmo: I like it because it leaves no unsightly artifacts in the codebase for things that are really up to the programmer

lib/traject/psulib_marc.rb
```diff
@@ -56,3 +58,99 @@ def process_genre(record, fields)
...
+# Refactor of Traject::Macros::Marc21Semantics#marc_publication_date as the basic logic but check for 264|*1|c before
+# 260c.
+def process_publication_date(record)
+  return nil unless record.is_a? MARC::Record
+
+  field008 = Traject::MarcExtractor.cached('008').extract(record).first
+  marc_date_processor = MarcPubDateProcessor.new(field008)
+  pub_date = marc_date_processor.find_date
+
+  if marc_date_processor.find_date.nil?
```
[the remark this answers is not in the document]
cdmo: yup!

lib/traject/psulib_marc.rb
```diff
@@ -56,3 +58,99 @@ def process_genre(record, fields)
...
+      date_str = @date2_str.to_i != 0 ? @date2_str : @date1_str
+      resolve_date date_str
+    when 'q'
+      # Questionable
+      resolve_range
+    else
+      # Default case, just resolve the first date.
+      resolve_date @date1_str
+    end
+  end
```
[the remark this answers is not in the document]
cdmo: thanks!

lib/traject/psulib_marc.rb
```diff
@@ -56,3 +58,99 @@ def process_genre(record, fields)
...
+
+  # For when we are dealing with ranges.
+  def resolve_range
+    # Make unknown digits at the beginning or end of range
+    date1 = @date1_str.sub('u', '0').to_i
+    date2 = @date2_str.sub('u', '9').to_i
+    # Do we have a range we can use?
+    is_range = (date2 > date1) && ((date2 - date1) <= ESTIMATE_TOLERANCE)
+    (date2 + date1) / 2 if is_range
```
banukutlu: @cdmo I added two more specs for `date_type=q`, one is failing. I want to see what you think about that case.

cdmo: 🤔 

```
(byebug) @date1_str
"19uu"
(byebug) @date1_str.sub('u', '0')
"190u"
```

cdmo: also, it's returning `'2002'` because the `@record` has a `260c`. Is that intentional? If it has a 260c, shouldn't we expect it to return that?

banukutlu: Yes, I agree it cannot find an acceptable date in 008, then it should return that 260c. But marc_publication_date is returning nil interestingly 🤔 I think it does not look correct that way, right?

lib/traject/psulib_marc.rb
```diff
@@ -126,11 +126,13 @@ def find_date
   # For when we are dealing with ranges.
   def resolve_range
     # Make unknown digits at the beginning or end of range
-    date1 = @date1_str.sub('u', '0').to_i
-    date2 = @date2_str.sub('u', '9').to_i
+    date1 = @date1_str.tr('u', '0').to_i
+    date2 = @date2_str.tr('u', '9').to_i
```
banukutlu: 👍

lib/traject/psulib_marc.rb
cdmo: `marc_publication_date`? That's gone, did you mean `marc_date_processor.find_date`? That should return nil. But then the next step finds the 260c and returns that value, which is 2002, and the test is expecting nil.

banukutlu: I was referring to  Traject's marc_publication_date which is using `Traject::Macros::Marc21Semantics.publication_date`. And yeah we are not using that anymore but I just tested against that and wanted to compare results.

cdmo: Yup yup I gotchya now, Slack convo cleared this up.

banukutlu: Per our conversation on slack, I will change this spec to test with a fixture that does not have a 264 neither 260. And other than that I think our logic for this case is working correctly - meaning it resorts to 264 or 260 if there is one when there is no date found in 008 due to ESTIMATE_TOLERANCE check failing.

banukutlu:
@cdmo it looks good to me now, what do you think?

cdmo:
:shipit:

Issue #83, opened 2019-01-23
Bound withs - Add parent info to generated 591 to display


banukutlu:
The display of the "Bound in" tag at the title level is misleading without item info.
eg. catkey [485144](https://blackcat01qa.libraries.psu.edu/catalog/485144). This title has 289 items under it but only one item is bound with another title(catkey [11749683](https://blackcat01qa.libraries.psu.edu/catalog/11749683)).

Add `949a`, `949c`, `949t` to generated 591s:

`591 a|Parent title c| Parent catkey t| Parent format n| Parent call number  c. copy number` 

so to display as below:

949a 949c (949t) bound in [title], e.g.

`AY67.N5W7 1922-24 c. 1 (MICROFORM) bound in The World almanac and encyclopedia`

or

`Microfilm D285 reel.478 no.3575 c.1 (MICROFORM) bound in The high-caste Hindu woman / With introduction by Rachel L. Bodley`

Also use translation map for `591t` format.

_Originally posted by @banukutlu in https://github.com/psu-libraries/psulib_traject/issues/42#issuecomment-456944505_

mkutch:
I noticed another issue. See https://blackcat01qa.libraries.psu.edu/catalog/988373. This has a bound with item but isn't showing up. I noticed the librarian view doesn't list all the 949s nor the 591. I went back to the extract and saw this record was broken down into 8 marc records (by catalogdump). Since the 591 build happens after the catalogdump it is on the 8th marc record. Checked with the extract and it looks like the librarian view is showing only the 7th marc record.
I can show you what I mean if this sounds confusing.

ruthtillman:
To-dos from 1/29 meeting:

@banukutlu - run script to combine with the mhlds and see how these a) index, b) combine in the record (put resulting combined MARC record and index debug result in Box?)

@mkutch see how/whether Stanford handles it when there's a number of libraries. Are they grouped by library in display? We may need to look at searchworks code / ask Shelley or a front-end dev how they're managed.

me - review MARC fields 853, 863, 867, 868, check in with John Hamilton. Check IU display. View web services result for these records.

ruthtillman:
[...]

PR #91, opened 2019-02-05
Adding properties for possibly bound in

ruthtillman:
Translation properties file for format type to a term which should be
used in the bound-in field. Note, this is a simplified form meant to be
a parenthetical and not necessarily used elsewhere.


cdmo:
this is merged into the feature branch `#83-bound-with-parent-info`

Issue #15, opened 2018-10-03
Mapping note fields


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Note-Fields-Index-and-Display

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 09:24

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 09:55

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 11:12

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 11:21

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 11:22

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 11:29

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 21, 2018, 13:19

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 22, 2018, 09:34

changed the description

banukutlu:
In GitLab by @rkt6 on Aug 22, 2018, 11:33

changed the description

banukutlu:
In GitLab by @rkt6 on Sep 4, 2018, 15:50

assigned to @rkt6

banukutlu:
In GitLab by @rkt6 on Sep 4, 2018, 15:50

assigned to @acg44 and unassigned @rkt6

banukutlu:
In GitLab by @acg44 on Sep 6, 2018, 11:04

changed the description

banukutlu:
In GitLab by @acg44 on Sep 6, 2018, 11:07

changed the description

banukutlu:
In GitLab by @acg44 on Sep 6, 2018, 11:21

changed the description

banukutlu:
In GitLab by @acg44 on Oct 3, 2018, 10:24

changed the description

banukutlu:
In GitLab by @acg44 on Oct 3, 2018, 11:27

changed the description

banukutlu:
In GitLab by @rkt6 on Oct 3, 2018, 12:03

Ruth - review 510s because they're a bit weird.

banukutlu:
[...]

PR #95, opened 2019-02-08
#15 mapping notes

cdmo:
Connected to #15

### DIFF
--- a/lib/traject/psulib_config.rb
+++ b/lib/traject/psulib_config.rb
@@ -405,8 +405,8 @@ to_field 'file_data_type_ssm', extract_marc('516a')
 # 518 - Data/Place of Event
 to_field 'date_place_event_note_ssm', extract_marc('518adop3')
 
-# 520 - Abstract
-to_field 'abstract_ssim', extract_marc('520ab')
+# 520 - Notes Summary
+to_field 'notes_summary_ssim', extract_marc('520ab')
 
 # 521 - Audience
 to_field 'audience_notes_ssm', extract_marc('521| *|3ab:521|8*|3ab:521|3*|3ab:521|4*|3ab')
--- a/lib/traject/psulib_config.rb
+++ b/lib/traject/psulib_config.rb
@@ -406,6 +406,7 @@ to_field 'file_data_type_ssm', extract_marc('516a')
 to_field 'date_place_event_note_ssm', extract_marc('518adop3')
 
 # 520 - Abstract
+to_field 'abstract_ssim', extract_marc('520ab')
 
 # 521 - Audience
 to_field 'audience_notes_ssm', extract_marc('521| *|3ab:521|8*|3ab:521|3*|3ab:521|4*|3ab')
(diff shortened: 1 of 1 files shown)


lib/traject/psulib_config.rb
```diff
@@ -359,6 +359,143 @@
...
+to_field 'restrictions_access_note_ssm', extract_marc('506abcdefu')
+
+# 507 - scale note for graphic material
+to_field 'scale_graphic_material_note_ssm', extract_marc('507ab')
+
+# 508 - Creation/Production Credits Note
+to_field 'creation_production_credits_ssm', extract_marc('508a')
+
+# 510 - Citation/References Note
+to_field 'citation_references_note_ssm', extract_marc('510| *|3abcux')
```
banukutlu: @cdmo `510| *|3abcux` is the space before * intentional?

lib/traject/psulib_config.rb
```diff
@@ -359,6 +359,143 @@
...
+to_field 'file_data_type_ssm', extract_marc('516a')
+
+# 518 - Data/Place of Event
+to_field 'date_place_event_note_ssm', extract_marc('518adop3')
+
+# 520 - Notes Summary
+to_field 'notes_summary_ssim', extract_marc('520ab')
+
+# 521 - Audience
+to_field 'audience_notes_ssm', extract_marc('521| *|3ab:521|8*|3ab:521|3*|3ab:521|4*|3ab')
```
banukutlu: @cdmo space here as well: `521| *|3ab`, it is probably OK and I also wanted to see if it means anything?

lib/traject/psulib_config.rb
cdmo: We discussed on Slack some, but yes, the space is intentional, as in the first indicator should be a space. Followed by the `*` which means "I don't care what the second indicator is."

cdmo: here's a good example, I think, go to

/catalog/12757052.refworks_marc_txt

`521 8  "Black publication."`

which mataches

`extract_marc('521| *|3ab:521|8*|3ab:521|3*|3ab:521|4*|3ab')`

specifically this part: `521|8*|3ab` because the `8` is the first indicator and a space is there for the second, but the `*` means it doesn't care.

lib/traject/psulib_config.rb
cdmo: Yes, actually my example in the previous conversation is from `521`

lib/traject/psulib_config.rb
banukutlu: I think I am more curious about if we only had `extract_marc('521| *|3ab')` for this example, would it match to `521 8 "Black publication."` (assuming 8 is the first indicator here)

cdmo: One white space is not a wildcard, but I'll go ahead and check.

banukutlu: per our slack convo, we cleared this out! So, space will match to space. Thanks @cdmo!

Issue #86, opened 2019-01-25
Add a combining reader for split records


banukutlu:
Per Maryam
> I noticed another issue. See https://blackcat01qa.libraries.psu.edu/catalog/988373. This has a bound with item but isn't showing up. I noticed the librarian view doesn't list all the 949s nor the 591. I went back to the extract and saw this record was broken down into 8 marc records (by catalogdump). Since the 591 build happens after the catalogdump it is on the 8th marc record. Checked with the extract and it looks like the librarian view is showing only the 7th marc record.
I can show you what I mean if this sounds confusing.

_Originally posted by @mkutch in https://github.com/psu-libraries/psulib_traject/issues/83#issuecomment-456969097_

Requires #88

banukutlu:
@mkutch can you upload an example extract for a such record on box?

banukutlu:
Per Ruth 
> Talked with Kutchemeshgi, Maryam this morning about the issue of split records in extracts with a lot of holdings and contacted Stanford to see who there works on theirs. Dawn at NCSU hasn't tackled it yet. Also, Maryam in a follow-up to our conversation--actual holdings would be displayed from WebServices, so it's more an issue for the bound-withs and anything else which uses the 949s for indexing...which still makes it a big issue, but when looking at others' catalogs they'd still display all the appropriate holdings... it just wouldn't be in librarianview/in their indexing that relies on holdings

ruthtillman:
Emailed person at Stanford responsible for their Sirsi exports. Copied Maryam and Banu.

banukutlu:
Stanford's solution: They added a class that combines the records (https://github.com/sul-dlss/searchworks_traject_indexer/blob/master/lib/traject/readers/marc_combining_reader.rb). 

traject config: 
https://github.com/sul-dlss/searchworks_traject_indexer/blob/eab76f7c71bf1650189bf9bbeee580e4b19f5c40/lib/traject/config/sirsi_config.rb#L254-L267

mkutch:
@banukutlu @cdmo  I found this which looks like their code in solr marc before the moved/converted their code to traject? Their algorithm is documented too. 

https://github.com/solrmarc/stanford-solr-marc/blob/master/core/src/org/solrmarc/marc/CombineMultBibsMhldsReader.java

mkutch:
@banukutlu @cdmo  I found this which looks like their code in solr marc before they moved/converted their code to traject? Their algorithm is documented too and can be helpful. 

[...]

PR #96, opened 2019-02-13
#86 combining reader

banukutlu:
Connected to #86


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2019-02-04 15:38:42 -0500 using RuboCop version 0.63.1.
+# on 2019-02-13 16:04:43 -0500 using RuboCop version 0.63.1.
```
cdmo: Our rubocop_todo keeps growing 😆

.ruby-version
```diff
@@ -1 +1 @@
-jruby-9.2.0.0
+jruby-9.2.5.0
```
cdmo: 👍 

this covers that bug that our colleagues mentioned in their own jruby debugging

Gemfile.lock
```diff
@@ -138,4 +127,4 @@ DEPENDENCIES
   traject-marc4j_reader
 
 BUNDLED WITH
-   1.17.1
+   2.0.1
```
cdmo: ah, bundler major upgrade

README.md
```diff
@@ -13,7 +13,7 @@ $ java --version
 Follow these instructions to [install JRuby](https://github.com/psu-libraries/psulib_traject/wiki/Install-JRuby) if you do not already have it.
 ```
 $ ruby --version
-  jruby 9.2.0.0
+  jruby 9.2.5.0
```
cdmo: 👌 nice catch!

lib/traject/psulib_config.rb
```diff
@@ -26,9 +27,9 @@
   provide 'log.file', 'log/traject.log'
   provide 'log.error_file', 'log/traject_error.log'
   provide 'solr_writer.commit_on_close', 'true'
+  provide 'reader_class_name', 'Traject::MarcCombiningReader'
```
cdmo: So, `Traject::MarcCombiningReader` requires `jruby`, and it even calls `marc4j_reader` itself. I would guess then, that extractions would fail when using MRI. However, I'm testing now and the transformation process seems to be running just fine. So... 🤷‍♂️ 

I guess my point was that we might want to move this reader class in the is_jruby conditional block.

Also, the MarcCombiningReader class looks for `JRUBY_VERSION` -- we should define that, right?

lib/traject/readers/marc_combining_reader.rb
```diff
@@ -0,0 +1,93 @@
+module Traject
```
cdmo: btw I mentioned that we are using this to Chris Beer at code4lib, told him thanks :)

spec/lib/traject/readers/marc_combining_reader_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      asplit1 = results.find { |r| r['001'].value == 'asplit1' }
+      expect(asplit1.fields('008').length).to eq 1
+      expect(asplit1.fields('245').length).to eq 1
+      expect(asplit1.fields('949').length).to eq 5
+      expect(asplit1.fields('949').map { |x| x['a'] }).to eq ['A1 .B2 V.1', 'A1 .B2 V.2', 'A1 .B2 V.3', 'A1 .B2 V.4', 'A1 .B2 V.5']
+
+      asplit3 = results.find { |r| r['001'].value == 'asplit3' }
+      expect(asplit3.fields('591').length).to eq 1
+      expect(asplit3.fields('591').map { |x| x['a'] }).to eq ['Parent Title']
```
cdmo: It's interesting, this is almost acting like an integration test for Maryam's extraction work, because she assembles these values on-the-fly during extraction.

Issue #97, opened 2019-02-14
Media types mapping


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Media-Types

banukutlu:
@ruthtillman are we gonna have a default value for the ones that are not caught by our rules

ruthtillman:
No, I think in cases where format and media type match -- e.g. a book that's a physical book ... we'll not add on. But if the book is a microfilm, etc. For some it'll always have a media type, like Video.

banukutlu:
@mkutch could you pls extract 21588551 for me?

PR #116, opened 2019-04-01
#97 media types

banukutlu:
Connected to #97 

Note: unrelated change to media types mapping: moved `@formats.compact!` and `@formats.uniq!` out of psulib_config.rb and added to marc_format_processor.rb


.rubocop_todo.yml
```diff
@@ -1,26 +1,31 @@
...
 
-# Offense count: 3
+# Offense count: 2
+# Configuration parameters: CountComments.
+Metrics/ClassLength:
+  Max: 140
+
+# Offense count: 4
 Metrics/CyclomaticComplexity:
-  Max: 7
```
banukutlu: :(

.rubocop_todo.yml
```diff
@@ -1,26 +1,31 @@
...
+  Max: 11
```
banukutlu: :(

banukutlu: I could not come up with a easy refactor, it could use another set of eyes :)

cdmo: Let's make sure to circle back to this one. I don't have a refactor for this one except for pushing every if statement embedded in whens into their own methods.

lib/marc_media_type_processor.rb
```diff
@@ -0,0 +1,181 @@
+# frozen_string_literal: true
+
+# A tool for classifying MARC records using a combination of data from the
+# record's leader and some 949ts to assign media types to records
+class MarcMediaTypeProcessor
```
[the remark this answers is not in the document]
cdmo: Once we have more classes, we can think about this maybe. Just a thought that is starting to form in my head anyway.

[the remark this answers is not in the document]
banukutlu: very good point, I really the parent class `MarcProcessor` idea 👍

PR #117, opened 2019-04-02
Docvalues

cdmo:
Coupled with https://github.com/psu-libraries/psulib_blacklight/pull/255


lib/traject/psulib_config.rb
```diff
@@ -171,10 +171,10 @@
 to_field 'author_ssort', marc_sortable_author
 
 ## Access facet
-to_field 'access_facet_ssim', extract_access_data
+to_field 'access_facet', extract_access_data
 
 # Formats and Resources
-to_field 'format' do |record, accumulator|
+to_field 'format' do |record, accumulator| #
```
banukutlu: as @cdmo describes, there is a stray octothorpe :)

cdmo: 😅

Issue #119, opened 2019-04-03
Refactor psulib_marc.rb


banukutlu:
- [x] move subject and genre work out of psulib_marc.rb
- [x] get rid of process_formats and move specs to its own formats specs
- [x] refactor process_publication_date and MarcPubDateProcessor should be thread-safe

cdmo:
So should this go to the backlog again because 2/3 of the tasks are done?

banukutlu:
> So should this go to the backlog again because 2/3 of the tasks are done?

yup 👍

PR #128, opened 2019-04-18
Moved subjects and genre work out of psulib_marc.rb and specs as well

banukutlu:
Connected to #119


lib/traject/macros/custom.rb
```diff
@@ -1,9 +1,82 @@
...
+              include_genre = vocabulary.include?(field['2'].to_s.downcase) if (field.tag == '655') && (field.indicator2 == '7')
+              genres << Traject::Macros::Marc21.trim_punctuation(genre) if include_genre
+            end
+          end
+
+          accumulator.replace(genres).uniq!
+        end
+      end
+
       NOT_FULLTEXT = /addendum|appendices|appendix|appendixes|cover|excerpt|executive summary|index/i.freeze
```
cdmo: I wonder if this constant should be put up on line 3 🤔 just for consistency

cdmo: (I know this is unrelated)

banukutlu: ahaa yeah makes sense

PR #130, opened 2019-04-19
98 urls

cdmo:
**Additional work**: Moved the NOT_FULLTEXT constant up


lib/traject/macros/custom.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 SEPARATOR = '—'.freeze
+NOT_FULLTEXT = /addendum|appendices|appendix|appendixes|cover|excerpt|executive summary|index/i.freeze
```
banukutlu: cool!

lib/traject/macros/custom.rb
```diff
@@ -87,11 +86,10 @@ def extract_link_data(link_type: 'full')
           link_data = []
 
           record.fields('856').each do |field|
-            break unless sought_link_data_exists?(link_type, field)
+            next unless sought_link_data_exists?(link_type, field)
```
banukutlu: 👍 good catch

Issue #32, opened 2018-10-03
Call Numbers mapping


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Call-Number-Index-and-Display

banukutlu:
In GitLab by @bzk60 on Sep 7, 2018, 15:44

added ~936 label

banukutlu:
In GitLab by @bzk60 on Sep 10, 2018, 10:07

changed the description

PR #131, opened 2019-04-20
mapping call numbers #32

banukutlu:
Connected to #32


banukutlu:
> Can I add subfield `a` to the wiki page? We just list `050` but don't mention the subfield which we extract data from. Or am I reading it wrong?

yes pls feel free to update/edit. thanks for catching it, I forgot about updating the wiki ;)

banukutlu:
> I sorta don't like the double `if`, but, it's simply not worth agonizing over. I do love how focused and small this is, easier to review :)

oh yeah double `if` :( I should have refactored that. But will keep in mind to do so next time I touch that file ;)

cdmo:
it could be the best solution, just kinda has a whiff :)

PR #135, opened 2019-05-01
63 incrementals

cdmo:
Note that I did add two new gems called `mail` and `whenever`

Additional changes: ran a bundle update which looks like it simply updated rubcop. I think that's it.

I didn't add any specs, just to be clear. After refactoring a bit and getting rid of some code it seemed simpler and less of a need for a spec. But, maybe I should still. I'd like your opinion, of course.

I do need to test on the blackcat01qa server too to see if it's *really* working.


Gemfile
```diff
@@ -4,11 +4,13 @@ source 'https://rubygems.org'
 
 gem 'coveralls', require: false
 gem 'library_stdnums'
+gem 'mail'
 gem 'marc'
 gem 'rake'
 gem 'rsolr'
-gem 'traject', '3.0.0'
+gem 'traject', '3.1.0'
```
banukutlu: 🆒 👍

lib/tasks/incrementals.rake
```diff
@@ -0,0 +1,82 @@
...
+      File.open(file_name, 'r') do |file|
+        file.each_line do |line|
+          id = line.chomp.chomp('|')
+          indexer.writer.delete(id)
+          response = HTTP.get "#{SOLR_URL}/select?defType=edismax&fq=id:#{id}"
+          parsed_response = JSON.parse(response)
+
+          # Sanity checking, may be uneccesary bloat because the SolrJsonWrite::delete will throw an exception if the
+          # response from the server isn't 200, and that only happens when the delete fails (as far as I know).
+          if parsed_response['response']['numFound'] != 0
```
banukutlu: 👍

lib/tasks/incrementals.rake
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+require 'mail'
+
+TRAJECT_HOME = '/opt/psulib_traject'.freeze
+TRAJECT_LOGS_HOME = '/var/log/traject'.freeze
+SIRSI_DATA_HOME = '/data/symphony_data'.freeze
+SOLR_URL = 'http://localhost:8983/solr/blacklight-core'.freeze
```
banukutlu: @cdmo can we use `indexer_settings['solr_url']` for this one as well?

banukutlu: oh maybe we do not need it anymore?

lib/tasks/incrementals.rake
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+require 'mail'
+
+TRAJECT_HOME = '/opt/psulib_traject'.freeze
+TRAJECT_LOGS_HOME = '/var/log/traject'.freeze
+SIRSI_DATA_HOME = '/data/symphony_data'.freeze
+
+# This job, and :delete_daily expect there to be file to add and delete, it is not responsible for getting those files
+# from the catalog.
```
banukutlu: thanks for noting this here 👍

banukutlu:
> idn't add any specs, just to be clear. After refactoring a bit and getting rid of some code it seemed simpler and less of a need for a spec. But, maybe I should s



> Note that I did add two new gems called `mail` and `whenever`
> 
> Additional changes: ran a bundle update which looks like it simply updated rubcop. I think that's it.
> 
> I didn't add any specs, just to be clear. After refactoring a bit and getting rid of some code it seemed simpler and less of a need for a spec. But, maybe I should still. I'd like your opinion, of course.
> 
> I do need to test on the blackcat01qa server too to see if it's _really_ working.

I agree that the tasks are simple enough, mainly running traject cmd and deleting files or sending an email if failed. I am fine with not adding specs for them yet. Maybe we can add an issue for backlog?

lib/tasks/incrementals.rake
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+    didnt_work = []
+    daily_deletion_files = Dir["#{SIRSI_DATA_HOME}/daily/*.txt"]
+    dont_delete = []
+
+    daily_deletion_files.each do |file_name|
+      File.open(file_name, 'r') do |file|
+        file.each_line do |line|
+          id = line.chomp.chomp('|')
```
banukutlu: @cdmo we do not need `chomp('|')` anymore, right?

banukutlu: @cdmo we do not need `chomp('|')` anymore, right?

cdmo: no more extra chomp :(

lib/tasks/incrementals.rake
```diff
@@ -53,7 +53,7 @@ namespace :incrementals do
     daily_deletion_files.each do |file_name|
       File.open(file_name, 'r') do |file|
         file.each_line do |line|
-          id = line.chomp.chomp('|')
+          id = line.chomp
```
banukutlu: yup no chomp chomp anymore 😞

lib/tasks/incrementals.rake
```diff
@@ -0,0 +1,86 @@
...
+
+# This job, and :delete_daily expect there to be file to add and delete, it is not responsible for getting those files
+# from the catalog.
+namespace :incrementals do
+  desc 'Adds to the index'
+  task :import_daily do
+    daily_addition_files = Dir["#{SIRSI_DATA_HOME}/daily/*.mrc"]
+
+    daily_addition_files.each do |f|
+      system("bundle exec traject -s log.file=#{TRAJECT_LOGS_HOME}/traject_incremental.log -s "\
```
banukutlu: @cdmo I am asking the same question (sorry), why do we have to do each here but not let traject cmd to run all the incrementals? 
`bundle exec traject ... -c #{TRAJECT_HOME}/lib/traject/psulib_config.rb #{SIRSI_DATA_HOME}/daily/*.mrc`

cdmo: To find ones that fail is what I was thinking. And if there's just one bad one, then at least you're processing the other stuff. But, yeah, it would be simpler to just do it this way. Chances are we'll catch the error quickly and fix it. I think your suggestion is the better solution. This has been merged already, so we can a quickfix branch kind of thing

lib/tasks/incrementals.rake
```diff
@@ -1,24 +1,24 @@
 # frozen_string_literal: true
 
 require 'mail'
+require 'traject'
 
-TRAJECT_HOME = '/opt/psulib_traject'.freeze
-TRAJECT_LOGS_HOME = '/var/log/traject'.freeze
```
banukutlu: 🎉

lib/tasks/incrementals.rake
```diff
@@ -1,24 +1,24 @@
...
 SIRSI_DATA_HOME = '/data/symphony_data'.freeze
 
 # This job, and :delete_daily expect there to be file to add and delete, it is not responsible for getting those files
 # from the catalog.
 namespace :incrementals do
   desc 'Adds to the index'
   task :import_daily do
+    indexer = Traject::Indexer::MarcIndexer.new
+    indexer.load_config_file('lib/traject/psulib_config.rb')
```
banukutlu: :cool: 👌

PR #137, opened 2019-05-01
134 regex fixes

cdmo:
Not sure if my fix was absolutely needed because it's a block (not a proc) inside a method, but, maybe better safe then sorry? I couldn't find any other issues, I looked for all `match`s and `split`s.


lib/traject/regex_split.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# work-around for https://github.com/jruby/jruby/issues/4868
```
banukutlu: @cdmo good call to move these to its own file 👍

cdmo: yeah I kind of had to do it because including `psulib_config.rb` into the marc processing class was a no go.

PR #145, opened 2019-05-09
Wrong indexer settings specified

cdmo:


lib/tasks/incrementals.rake
```diff
@@ -32,7 +32,7 @@ namespace :incrementals do
   desc 'Deletes from the index'
   task :delete_daily do
     require 'yaml'
-    indexer_settings = YAML.load_file('config/indexer_settings.yml')
+    indexer_settings = YAML.load_file('config/indexer_settings_production.yml')
```
banukutlu: ahaa! 👌

cdmo: oops :)

PR #150, opened 2019-05-15
* Adds ability to specify qa or prod

cdmo:
* Adds ability to specify hourly or daily

Just allowed the passing of an arg for period and modified the tasks to account for that. This is assuming things about where the data is put. This is subject to change when the production indexer server is ready

### DIFF
--- a/lib/tasks/incrementals.rake
+++ b/lib/tasks/incrementals.rake
@@ -8,13 +8,13 @@ SIRSI_DATA_HOME = '/data/symphony_data'.freeze
 # from the catalog.
 namespace :incrementals do
   desc 'Adds to the index'
-  task :import_daily do
+  task :import, [:period] do |_task, args|
     require 'mail'
     today_ymd = Date.today.strftime('%Y%m%d')
-    file = "#{SIRSI_DATA_HOME}/daily/daily_addupdate_#{today_ymd}.mrc"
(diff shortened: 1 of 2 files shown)


lib/tasks/incrementals.rake
```diff
@@ -8,13 +8,13 @@ SIRSI_DATA_HOME = '/data/symphony_data'.freeze
...
   desc 'Adds to the index'
-  task :import_daily do
+  task :import, [:period] do |_task, args|
     require 'mail'
     today_ymd = Date.today.strftime('%Y%m%d')
-    file = "#{SIRSI_DATA_HOME}/daily/daily_addupdate_#{today_ymd}.mrc"
     indexer = Traject::Indexer::MarcIndexer.new
     indexer.load_config_file('lib/traject/psulib_config.rb')
-    indexer.logger.info "   Processing incremental import_daily rake task on #{file}"
+    file = "#{SIRSI_DATA_HOME}/#{args[:period]}_#{psulib_resolve_environment}/daily_addupdate_#{today_ymd}.mrc"
```
banukutlu: @cdmo `daily_addupdate_#{today_ymd}.mrc` should the daily part change as `#{args[:period]}`? 🤔 
how are the hourly files named?

cdmo: gah, i even double checked, but yeah, good catch

Issue #172, opened 2019-07-09
Fixing one issue with Archives/Manuscripts


ruthtillman:
One of the reasons "regular" books are ending up in the Archives/Manuscripts area is that a few campuses are putting the 949t as ARCHIVES when they keep a copy with their archival materials. However, the leader(6-7) is still "am" not "tm." Other copies may be in regular shelves at other locations. Based on the number of books, it looks like the most efficient fix would be to prioritize the leader "am" over 949t = ARCHIVES.

I don't know if there's a good way to say:

```
if the leader is "am"
AND a 949t is ARCHIVES
use the "am" mapping, not the ARCHIVES mapping
```

banukutlu:
@ruthtillman do you have any example catkeys?

ruthtillman:
Sample catkeys with this issue: 

- 1968855
- 1586824
- 20821274
- 17742510

(these include a variety of book types, but even books about archives should just be Books)

PR #182, opened 2019-07-22
added an override to formats mapping for book - should prefer leader(…

banukutlu:
…6-7) mapping over 949t

Issue #172


cdmo:
This will require a full extract.

PR #191, opened 2019-08-22
Bundle update to bump nokogiri because of security issue

cdmo:


cdmo:
super weird that it's failing CI. This PR doesn't change any ruby code

Issue #167, opened 2019-06-26
Index specialized sublocation for Aeon


ruthtillman:
We will need to index 799a for the sublocation to be passed to Aeon. Sample record: c107

@mkutch needs to first see if this is being extracted or not because it doesn't seem to be in our MARC exports.
 
OR, should we use the staff notes if sending to Aeon? See the staff notes on the response, which would have the right location for each particular item https://cat.libraries.psu.edu:28443/symwsbc/rest/standard/lookupTitleInfo?clientID=BCCAT&titleID=107&includeAvailabilityInfo=true&includeItemInfo=true

this seems more accurate

ruthtillman:
Re: the staff note, we'd just need to test a lot to be sure that they're the kind of data we want.

mkutch:
@ruthtillman I've been reading the webservices documentation and there is another webservice variable we can add and turn at end of url .
It is marcEntryFilter and it has values ALL, BRIEF, FULL, NONE, TEMPLATE
This will give us the MARC data! Each value refers to what has been set to display in the policy (we see it when we do a Change Display in the CAT).

https://cat.libraries.psu.edu:9443/symwsbc/rest/standard/lookupTitleInfo?clientID=BCPAC&titleID=107&includeAvailabilityInfo=true&includeItemInfo=true&includeBoundTogether=true&marcEntryFilter=ALL

and that's not all. We can even specify a specific field we want.

https://cat.libraries.psu.edu:9443/symwsbc/rest/standard/lookupTitleInfo?clientID=BCPAC&titleID=107&includeAvailabilityInfo=true&includeItemInfo=true&includeBoundTogether=true&marcEntryID=799

We can even list the fields we want to return in webservice call so we won't have to index fields if they aren't necessary.

https://cat.libraries.psu.edu:9443/symwsbc/rest/standard/lookupTitleInfo?clientID=BCPAC&titleID=107&includeAvailabilityInfo=true&includeItemInfo=true&includeBoundTogether=true&marcEntryID=245,799,506,264

mkutch:
@ruthtillman It’s best not to use Staff notes field in this case. Staff notes can contain anything (i.e. Hathitrust notes, Digitization notes, etc)

ruthtillman:
[...]

PR #197, opened 2019-10-21
mapping 799a sublocation field

banukutlu:
Issue #167 and also additional work: Issue #199

### DIFF
--- a/lib/traject/psulib_config.rb
+++ b/lib/traject/psulib_config.rb
@@ -144,7 +144,6 @@ to_field 'title_related_tsim', extract_marc(%w[
   796lktmnoprs
   797lktmnoprs
   798lktmnoprs
-  799alktmnoprs
 ].join(':'), trim_punctuation: true) do |record, accumulator|
   accumulator.each { |value| value.chomp!(' --') } unless record.fields('505').empty?
 end
--- a/lib/traject/psulib_config.rb
+++ b/lib/traject/psulib_config.rb
@@ -568,4 +568,6 @@ to_field 'serials_merged_to_form_display_ssim', extract_marc('785|07|iabdghkmnop
 to_field 'serials_changed_back_to_display_ssim', extract_marc('785|08|iabdghkmnopqrstuxyz3')
 
 # 799a - Sublocation
+# From our catalog experts: "The data in this field corresponds to "collections" within Special Collections, and 
+# the 799 data lets their staff know which shelf (or range of shelves) to check for the call number in question.
 to_field 'sublocation_ssm', extract_marc('799a')
--- a/lib/traject/psulib_config.rb
+++ b/lib/traject/psulib_config.rb
@@ -566,3 +566,6 @@ to_field 'serials_separated_from_display_ssim', extract_marc('780|07|iabdghkmnop
 to_field 'serials_split_into_display_ssim', extract_marc('785|06|iabdghkmnopqrstuxyz3')
 to_field 'serials_merged_to_form_display_ssim', extract_marc('785|07|iabdghkmnopqrstuxyz3')
 to_field 'serials_changed_back_to_display_ssim', extract_marc('785|08|iabdghkmnopqrstuxyz3')
+
+# 799a - Sublocation
+to_field 'sublocation_ssm', extract_marc('799a')
--- a/lib/traject/psulib_config.rb
+++ b/lib/traject/psulib_config.rb
@@ -568,6 +568,6 @@ to_field 'serials_merged_to_form_display_ssim', extract_marc('785|07|iabdghkmnop
 to_field 'serials_changed_back_to_display_ssim', extract_marc('785|08|iabdghkmnopqrstuxyz3')
 
 # 799a - Sublocation
-# From our catalog experts: "The data in this field corresponds to "collections" within Special Collections, and 
+# From our catalog experts: "The data in this field corresponds to "collections" within Special Collections, and
 # the 799 data lets their staff know which shelf (or range of shelves) to check for the call number in question.
 to_field 'sublocation_ssm', extract_marc('799a')


cdmo:
Also, should there be more logic to this? Something to test if a thing is an AEON item in the first place? Because:

```
43325        sublocation_ssm           Jean Giraudoux
49547        sublocation_ssm           Ramón del Valle-Inclán
51301        sublocation_ssm           Theophrastus
53131        sublocation_ssm           Westvaco Press
53716        sublocation_ssm           Henry Shoemaker
54380        sublocation_ssm           Something Else Press
54419        sublocation_ssm           Something Else Press
54429        sublocation_ssm           Christmas
54431        sublocation_ssm           Black Sparrow Press
54432        sublocation_ssm           Black Sparrow Press
54438        sublocation_ssm           Black Sparrow Press
54440        sublocation_ssm           Black Sparrow Press
54442        sublocation_ssm           Black Sparrow Press
54443        sublocation_ssm           Black Sparrow Press
54446        sublocation_ssm           Black Sparrow Press
54641        sublocation_ssm           Theodore Roethke
55052        sublocation_ssm           Samuel Johnson
55061        sublocation_ssm           Samuel Johnson
55593        sublocation_ssm           Samuel Johnson
55596        sublocation_ssm           Samuel Johnson
55602        sublocation_ssm           Samuel Johnson
56372        sublocation_ssm           Alexander Pope
56449        sublocation_ssm           John Dryden
56471        sublocation_ssm           John Dryden
56543        sublocation_ssm           Henry Fielding
56673        sublocation_ssm           Oliver Goldsmith
56675        sublocation_ssm           Oliver Goldsmith
56678        sublocation_ssm           Oliver Goldsmith
56715        sublocation_ssm           Oliver Goldsmith
56717        sublocation_ssm           Oliver Goldsmith
56719        sublocation_ssm           Oliver Goldsmith
56721        sublocation_ssm           Oliver Goldsmith
56723        sublocation_ssm           Oliver Goldsmith
56725        sublocation_ssm           Oliver Goldsmith
56764        sublocation_ssm           Charles Churchhill
56774        sublocation_ssm           Alexander Pope
56778        sublocation_ssm           Samuel Johnson
56877        sublocation_ssm           John Milton
56908        sublocation_ssm           William Cowper
56960        sublocation_ssm           John Milton
57051        sublocation_ssm           Samuel Johnson
57108        sublocation_ssm           Samuel Johnson
57181        sublocation_ssm           Samuel Richardson
57196        sublocation_ssm           Jonathan Swift
57255        sublocation_ssm           James Thomson
57258        sublocation_ssm           James Thomson
57275        sublocation_ssm           James Thomson
57548        sublocation_ssm           Alexander Pope
57595        sublocation_ssm           Tobias Smollett
57598        sublocation_ssm           Tobias Smollett
57600        sublocation_ssm           Tobias Smollett
58162        sublocation_ssm           Marie Corelli
```

Seems odd to store this info?

banukutlu:
> 1. From [wiki page](https://github.com/psu-libraries/psulib_blacklight/wiki/Requesting-Items-with-Aeon#aeon-mappings-for-archival-materials):
> 
> > can be multivalued and joined with separator `;`
> 
> So don't you have to add `separator: ';'`?
> 
> 1. Also, the `799a` is already mapped to `title_related_tsim` which makes sense I guess as the [MARC 21 docs](https://www.oclc.org/bibformats/en/7xx/799.html) say this is supposed to be uniform title. Is that a problem?
> 2. Can you move this up to the Location chunk of `psulib_config.rb` that starts at L544?
> 3. What exactly is subLocation? Is it the same as location in the sense of like "circ desk"?

1. Only couple records supposed to have multiple values and I thought we would just map the values as is, and handle the separator when we display, if we decide to change it, would be easier but I also thought about doing it while mapping. So I do not have a strong opinion. 

2. & 3. This is not actually a location more like a category from what I understand.

cdmo:
> but I thought about doing it while mapping also so I do not have a strong opinion

I think it makes sense to do while mapping. Might as well store the data how we will display it especially since the ability to change separators is an easy option to throw in (built-in to traject). That's my 2 cents. Are there any instances where you wouldn't want the semicolon?

banukutlu:
> > but I thought about doing it while mapping also so I do not have a strong opinion
> 
> I think it makes sense to do while mapping. Might as well store the data how we will display it especially since the ability to change separators is an easy option to throw in (built-in to traject). That's my 2 cents. Are there any instances where you wouldn't want the semicolon?

I thought since they are more like categories, I would index them w/o a separator. But since as of now it is not a field to be searched on it makes sense to store it the way we want to display. So I will add the separator and change the name of the field as `sublocation_display_ssm`.
Yes, changing the separator is always easy just needs a full reindexing. But I do not think that we will need to change it for this field. So we are good. If there would be a need to search on this field, I guess we would add another field which is indexed.

cdmo:
Sounds good. It's interesting the "`display`" part of the field name, I hadn't considered that as a naming pattern/convention we were using, but yep, checks out that we are.

lib/traject/psulib_config.rb
```diff
@@ -566,3 +566,6 @@
 to_field 'serials_split_into_display_ssim', extract_marc('785|06|iabdghkmnopqrstuxyz3')
 to_field 'serials_merged_to_form_display_ssim', extract_marc('785|07|iabdghkmnopqrstuxyz3')
 to_field 'serials_changed_back_to_display_ssim', extract_marc('785|08|iabdghkmnopqrstuxyz3')
+
+# 799a - Sublocation
```
cdmo:
```suggestion
# 799a - Sublocation
# From our catalog experts: "The data in this field corresponds to "collections" within Special Collections, and 
# the 799 data lets their staff know which shelf (or range of shelves) to check for the call number in question.
```

PR #208, opened 2020-05-26
CodeClimate does not work with version 0.18 or greater of simplecov

cdmo:
See https://github.com/codeclimate/test-reporter/issues/418


cdmo:
Also see https://travis-ci.com/github/psu-libraries/psulib_traject/builds/167877594#L1351

PR #210, opened 2020-06-02
hathi etas tasks

banukutlu:
Issue #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ $ ruby --version
    
    You can download [other sample files from Box](https://psu.app.box.com/folder/53004724072).
    
-### Traject in debug mode
+## Traject in debug mode
 
 For testing purposes you can run traject with the `--debug-mode` flag to
 display the output to the console (and not push the data to Solr).
@@ -64,3 +64,25 @@ display the output to the console (and not push the data to Solr).
 ```
 $ bundle exec traject --debug-mode -c lib/traject/psulib_config.rb solr/sample_data/sample_psucat.mrc
 ```
+
+## Hathi Trust ETAS data
+
+To generate a synthesized Hathi Trust overlap report locally, you can run 
+
+```
+RUBY_ENVIRONMENT=dev bundle exec rake hathitrust:process_hathi_etas`.
+```
+
+You will need the overlap file provided by Hathi Trust (`overlap_[date]_psu.tsv`) in your `ignorethis_hathi/` directory.
+
+Also make sure the below settings in your `indexer_settings_dev.yml` file are set with the correct info:
+
+
+```
+hathi_load_period: 20200501
+overlap_file: overlap_20200518_psu.tsv
+```
+
+The generated file is used during the import process to index `ht_id` by checking 
+on an existing match to OCLC number and also for "Access Online". For more info, see
+[Synthesizing-overlap-data-from-HathiTrust](https://github.com/psu-libraries/psulib_blacklight/wiki/Synthesizing-overlap-data-from-HathiTrust)
--- a/lib/tasks/hathi_etas.rake
+++ b/lib/tasks/hathi_etas.rake
@@ -5,7 +5,7 @@ HATHI_DATA_HOME = '/data/hathitrust_data'
 namespace :hathitrust do
   desc 'Process overlap file for emergency access to restricted HathiTrust material'
   task :process_hathi_etas, [:hathi_full, :overlap_psu] do |_task, args|
-    Dir.chdir("#{HATHI_DATA_HOME}") do
+    Dir.chdir(HATHI_DATA_HOME.to_s) do
       Rake::Task['hathitrust:pare_hathi_full'].invoke(args[:hathi_full])
       Rake::Task['hathitrust:extract_overlap_oclc'].invoke(args[:overlap_psu])
       Rake::Task['hathitrust:filter_overlap'].invoke
--- a/lib/tasks/hathi_etas.rake
+++ b/lib/tasks/hathi_etas.rake
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+HATHI_DATA_HOME = '/data/hathitrust_data'
+
+namespace :hathitrust do
+  desc 'Process overlap file for emergency access to restricted HathiTrust material'
+  task :process_hathi_etas, [:hathi_full, :overlap_psu] do |_task, args|
+    Dir.chdir("#{HATHI_DATA_HOME}") do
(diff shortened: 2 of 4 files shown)


cdmo:
wonders if these tools are installable via `yum` and if so we need to add them (preferably through our deploy project)

lib/tasks/hathi_etas.rake
```diff
@@ -0,0 +1,41 @@
...
+      Rake::Task['hathitrust:pare_hathi_full'].invoke(args[:hathi_full])
+      Rake::Task['hathitrust:extract_overlap_oclc'].invoke(args[:overlap_psu])
+      Rake::Task['hathitrust:filter_overlap'].invoke
+      Rake::Task['hathitrust:extract_uniq_oclc'].invoke
+    end
+  end
+
+  desc 'Pare the monthly file down to just the needed data: Hathi Trust id and OCLC number'
+  task :pare_hathi_full, [:hathi_full] do |_task, args|
+    print `gunzip -c #{args[:hathi_full]} | \
```
cdmo, line 18: should we just make it wget/curl the full data set? I think it has a specific pattern for file naming

lib/tasks/hathi_etas.rake
```diff
@@ -0,0 +1,41 @@
...
+               csvcut -t -c 1,8 -z 1310720 | \
+               csvgrep -c 1,2 -r ".+" | \
+               sort | uniq > hathi_full_dedupe.csv`
+  end
+
+  desc 'Extract the unique set of OCLC numbers from the overlap report'
+  task :extract_overlap_oclc, [:overlap_psu] do |_task, args|
+    print `csvgrep -t -c 4 -r ".+" #{args[:overlap_psu]} | \
```
cdmo, line 26: same here we could curl it. If we knew the filename on the HT server we could use that as part of a formula for the curl operation.

lib/tasks/hathi_etas.rake
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+HATHI_DATA_HOME = '/data/hathitrust_data'
```
cdmo, line 3: I am trying now to use the `settings` present in psulib_config.rb. Maybe you can load that here? I'm not 100% sure what the "right way" of doing this would be, but once loaded you could just use `settings['hathi_overlap_path']`

Something like https://github.com/psu-libraries/psulib_traject/blob/master/lib/tasks/incrementals.rake#L13-L14 would have to be run first, then `indexer.settings` could be accessed (maybe)

welcome other ideas

lib/tasks/hathi_etas.rake
banukutlu: > wonders if these tools are installable via `yum` and if so we need to add them (preferably through our deploy project)

that's what I was kinda hoping so as well that we can install them with our deploy process

lib/tasks/hathi_etas.rake
banukutlu: yes, I think so, I will give it a try

lib/tasks/hathi_etas.rake
banukutlu: yes, I think so, I will give it a try

banukutlu:
> wonders if these tools are installable via `yum` and if so we need to add them (preferably through our deploy project)

that's what I was kinda hoping so as well that we can install them with our deploy process

cdmo:
Think we should move the deploy repo into psu-libraries? Now that we get unlimited free private repos it seems like a worthwhile effort, then everything is tracked together in the BlackCat project...

banukutlu:
> Think we should move the deploy repo into psu-libraries? Now that we get unlimited free private repos it seems like a worthwhile effort, then everything is tracked together in the BlackCat project...

yup, I was thinking the same thing!

lib/tasks/hathi_etas.rake
banukutlu: yes it makes sense to get it from indexer.settings, I will give it a try, thnx! 👍

banukutlu: this is done with the last commit

lib/tasks/hathi_etas.rake
banukutlu: added this

lib/tasks/hathi_etas.rake
banukutlu: not sure where the overlap file is on the HT side, there is the shared box folder but I do not know if we can find it elsewhere 🤔

cdmo: well lets ask a follow up question to HT when they write back to us

cdmo:
linked https://github.com/psu-libraries/psulib_blacklight_deploy/issues/75

lib/tasks/hathi_etas.rake
banukutlu: we talked about this on slack, we will handle moving the overlap file manually.

cdmo:
Ok, so how do I run it? `bundle exec rake hathitrust`?

banukutlu:
> Ok, so how do I run it? `bundle exec rake hathitrust`?

`RUBY_ENVIRONMENT=dev  bundle exec rake hathitrust:process_hathi_etas`
and also we will need the overlap file in `ignorethis_hathi/`

README.md
```diff
@@ -56,11 +56,33 @@ $ ruby --version
...
 ```
 $ bundle exec traject --debug-mode -c lib/traject/psulib_config.rb solr/sample_data/sample_psucat.mrc
 ```
+
+## Hathi Trust ETAS data
+
+To generate a synthesized Hathi Trust overlap report locally, you can run 
+
+```
+RUBY_ENVIRONMENT=dev bundle exec rake hathitrust:process_hathi_etas`.
```
cdmo, line 73:
```suggestion
RUBY_ENVIRONMENT=dev bundle exec rake hathitrust:process_hathi_etas --overlap_file=name_of_file.csv`
```

is this right?

README.md
```diff
@@ -56,11 +56,33 @@ $ ruby --version
...
+```
+
+You will need the overlap file provided by Hathi Trust (`overlap_[date]_psu.tsv`) in your `ignorethis_hathi/` directory.
+
+Also make sure the below settings in your `indexer_settings_dev.yml` file are set with the correct info:
+
```
cdmo, line 79:
```suggestion

Any of the settings in `indexer_setting` files can be overridden like any other traject config with a flag like `--overlap_file=some_special_file.csv`.
```

cdmo: nevermind, I misunderstood

Issue #204, opened 2020-05-18
Map HathiTrust overlap data


banukutlu:
Using the overlap file, index data into the Solr collection so faceting is available.

- Volume Identifier - `htid` 
- HathiTrust Record Number - `ht_bib_key`
- Access Code - `access`

## Include Hathi ETAS on/off toggle in `Settings`

When on: access facet is set to online for denies
When off: access facet is not set for denies

https://www.hathitrust.org/how-to-add-etas-records-to-your-catalog#HathiFiles

cdmo:
For what we are trying to do (add records to the access online facet) I don't think we need to add anything to the index, instead we can just do a lookup in the access facet field processing.

cdmo:
@banukutlu and I just chatted for a bit and confirmed we need htid and that will replace the need for the bibapi javascript call we make in blackcat

cdmo:
I'm merging your branch into mine @banukutlu, continue the work

cdmo:
The review of this is covered by #585, so I'm closing this one.

PR #215, opened 2020-06-10
hathitrust ingest enhancements

banukutlu:
Issue #204

##  Notes
Need to run `RUBY_ENVIRONMENT=dev bundle exec rake hathitrust:process_hathi_overlap` locally to generate required files for hathi data ingest

##  Todo 

- [x] ~~rows with multiple oclc's from hathi full file seems to be not matching. 
  eg. `"9679497,9679922,9680110",coo.31924094207804,009153399,allow~~ - this will be handled by #219
- [x] specs update
- [x] change to `hathitrust_struct` with id and access code and update `access` facet logic accordingly
- [x] add hathi etas on/off in settings


config/indexer_settings_dev.yml
```diff
@@ -2,6 +2,7 @@
 
 processing_thread_pool: 5
 hathi_overlap_path: ignorethis_hathi/
-hathi_overlap_file: final_hathi_overlap.csv
+hathi_mono_overlap_file: hathi_filtered_by_mono_overlap.csv
+hathi_multi_overlap_file: hathi_filtered_by_multi_overlap.csv
```
cdmo, line 6: thinking one file with `ht_id`, `ht_bib_key` and `ht_access` (for deny/allow) would be everything we need.

All multis will be linked to `ht_bib_key` in blackcat, but, we'll still need deny/allow to inform language of the link.

lib/traject/macros/custom.rb
```diff
@@ -205,14 +205,26 @@ def self.includes_oclc_indicators?(sf_a)
...
         end
       end
+
+      # Extract ht_id
+      def extract_ht_id
+        lambda do |_record, accumulator, context|
+          ht_bib_key = context.output_hash&.dig('ht_bib_key_ssim')&.first
+          if ht_bib_key.nil?
+            oclc_number = context.output_hash&.dig('oclc_number_ssim')&.first
+            accumulator << HATHI_MONO_OVERLAP&.dig(oclc_number, :ht_id)
```
cdmo, line 223: Would be nice for this method to be as simple as something like:

```rb
            oclc_number = context.output_hash&.dig('oclc_number_ssim')&.first
            accumulator << HATHI_OVERLAP&.dig(oclc_number, :ht_id)
```

banukutlu:
```
oclc,item_type,access
1038694,mono,allow
1038694,mono,deny
1038694,multi,allow
1038694,multi,deny
```
because there are records like above, needed to check if there was a `ht_bib_key` here since 1038694 will be in both files.
OR this check could be done while generating the final file to lookup like the final file would have 

I found it to be easier to do this way, but as we discussed, please feel free to give it a try for a combined final file, I might have not think of a sol'n you could come up with.

banukutlu: I think we already decided that this record would go to record page so ht_bib_key indexed. Since when dividing up the files I am not checking this logic, I am checking it when indexing ht_id. Sure, it can be checked during massaging the data so the final file would only have 
`1038694,multi,allow` for this record.

cdmo: So does that mean we're trying to make one final file with `oclc`, `ht_id`, `ht_bib_key`, `type` (i.e., `mono` or `multi`) and `access` be nice to deal with? You could use a single easy lookup for any thing and either `ht_id` or `ht_bib_key` would be blank so that record might end up looking like

`1038694,,[the_ht_bib_key_value],multi,allow`

I guess we just can't combine the csvs right now because csvjoin is too costly? Is that the problem? Or is having two files just fine and let's move on! :)

banukutlu: I mentioned the one file since I was thinking that is what you wanted to try. I think I'm personally ready to move on from playing around with the data :) But if you come with that one file, I would happily use it :)

banukutlu: cvsjoin is said to be costly for big files but I haven't tried with these two files tbh maybe it would be quick since they are smaller now that after removing columns out, It just did not seem necessary since looking up to one file for each field seemed fine. Also then why divide them up in the first place then if you are going with one file approach.

cdmo: yup okay with moving on with two files, all good.

config/indexer_settings_dev.yml
banukutlu: https://github.com/psu-libraries/psulib_traject/pull/215#discussion_r441220566 moving on with two file for now

lib/tasks/hathi_trust.rake
```diff
@@ -20,29 +20,42 @@ namespace :hathitrust do
     print `curl -s -O 'https://www.hathitrust.org/sites/www.hathitrust.org/files/hathifiles/hathi_full_#{args[:period]}.txt.gz'`
   end
 
-  desc 'Pare the monthly file down to just the needed data: Hathi Trust id and OCLC number'
+  desc 'Pare the monthly file down to just the needed data: OCLC number, Hathi Trust id, Hathi Trust bib_key and Hathi Turst access code'
```
cdmo, line 23:
```suggestion
  desc 'Pare the monthly file down to just the needed data: OCLC number, HathiTrust id, HathiTrust bib_key and HathiTurst access code'
```

config/indexer_settings_dev.yml
```diff
@@ -2,6 +2,8 @@
...
 processing_thread_pool: 5
 hathi_overlap_path: ignorethis_hathi/
-hathi_overlap_file: final_hathi_overlap.csv
+hathi_mono_overlap_file: final_hathi_mono_overlap.csv
+hathi_multi_overlap_file: final_hathi_multi_overlap.csv
 hathi_load_period: 20200501
-overlap_file: overlap_20200518_psu.tsv
\ No newline at end of file
+overlap_file: overlap_20200518_psu.tsv
+hathi_etas: true
```
cdmo, line 9: For indexer environment level configs I made a PR recently to kinda pare them down a lot. My main thinking is that it's easier to just manage config that is the same everywhere right in the `settings` block, and that's what I'm thinking about `hathi_etas` too.

config/indexer_settings_test.yml
```diff
@@ -2,4 +2,6 @@
 
 processing_thread_pool: 5
 hathi_overlap_path: spec/fixtures/hathitrust/
-hathi_overlap_file: mock_overlap.csv
\ No newline at end of file
+hathi_mono_overlap_file: mock_mono_overlap.csv
```
cdmo, line 5: Same thing for the files, I only ass values to env indexer setting files if they are different from the default.

lib/marc_access_facet_processor.rb
```diff
@@ -43,4 +43,12 @@ def resolve_library_code(field, library)
 
     'On Order'
   end
+
+  def hathi_access?(context)
+    hathitrust_struct = context.output_hash&.dig('hathitrust_struct')
+    hathitrust_etas = context.settings['hathi_etas']
+    return false unless hathitrust_struct
+
+    hathitrust_etas || (JSON.parse(hathitrust_struct&.first)['access'] == 'allow' && !hathitrust_etas)
```
cdmo, line 52:
```suggestion
    hathitrust_etas || JSON.parse(hathitrust_struct&.first)['access'] == 'allow'
```

Seems redundant to check `hathitrust_etas` twice. If `hathitrust_etas` is true, the part of this statement after `||` never gets evaluated. If it is false, the part after `||` will mean `!hathitrust_etas` always evaluates to true. Right?

banukutlu: yes. right 👍

lib/traject/macros/custom.rb
```diff
@@ -205,12 +205,22 @@ def self.includes_oclc_indicators?(sf_a)
...
         lambda do |_record, accumulator, context|
           oclc_number = context.output_hash&.dig('oclc_number_ssim')&.first
-          accumulator << HATHI_ETAS_OVERLAP[oclc_number]
-          accumulator.compact
+          ht_hash = HATHI_MULTI_OVERLAP&.dig(oclc_number)&.first || HATHI_MONO_OVERLAP&.dig(oclc_number)&.first
+          accumulator << ht_hash.to_json unless ht_hash.nil?
+        end
+      end
+
+      def hathi_to_hash(ht_format)
```
cdmo, line 216: Can you add a spec to cover this method?

I am really having a hard time with line 219 especially. It's using a ["pretzel colon"](https://www.honeybadger.io/blog/how-ruby-ampersand-colon-works/) to do the following:

"create a new array that is grouped by each value in the array with the first element removed"

Is that understanding correct?

banukutlu: That line will group the records by oclc by removing the oclc from the array:

```
# [[1, 10, allow] [1, 20, deny], [2, 30, allow], [2, 40, deny]]
result = result.group_by(&:shift)
# {1 => [[10, allow], [20, deny]], 2 => [[30, allow], [40, deny]]}
#  ^ oclc                          ^ oclc
```

banukutlu: I intended to have this method as private so I did not add a spec for it (I think the other specs kinda covering it) 🤔 
I just wasn't sure where the better place to put this method would be (here?), wanted to see what your thoughts are on it.

banukutlu: specs added for this method

config/indexer_settings_test.yml
banukutlu: 🤔 I added these because they are different from the default ones, so sorry I think I do not understand what change you request/recommend here?

banukutlu: maybe you meant to write this to the dev file? I can remove the ones in the dev file.

cdmo: oh I'm so sorry! I must have read that wrong. :(

README.md
```diff
@@ -65,15 +65,19 @@ display the output to the console (and not push the data to Solr).
...
 ```
 RUBY_ENVIRONMENT=dev bundle exec rake hathitrust:process_hathi_etas`.
 ```
 
-You will need the overlap file provided by Hathi Trust (`overlap_[date]_psu.tsv`) in your `ignorethis_hathi/` directory.
+You will need the overlap file provided by HathiTrust (`overlap_[date]_psu.tsv`) and `hathi_field_list.csv` with the content
+
+`oclc_num,htid,ht_bib_key,access`
+
+in your `ignorethis_hathi/` directory.
```
cdmo, line 80:
```suggestion
in your `ignorethis_hathi/` directory (or whatever directory path you set in `hathi_overlap_path`).
```

Just thinking that this would make it clear that it can be overridden? Especially in other environments?

lib/traject/psulib_config.rb
```diff
@@ -38,7 +38,9 @@
   provide 'reader_class_name', 'Traject::MarcCombiningReader'
   provide 'commit_timeout', '10000'
   provide 'hathi_overlap_path', indexer_settings['hathi_overlap_path'] || '/data/hathitrust_data/'
-  provide 'hathi_overlap_file', indexer_settings['hathi_overlap_file'] || 'final_hathi_overlap.csv'
+  provide 'hathi_mono_overlap_file', indexer_settings['hathi_mono_overlap_file'] || 'final_hathi_mono_overlap.csv'
+  provide 'hathi_multi_overlap_file', indexer_settings['hathi_multi_overlap_file'] || 'final_hathi_multi_overlap.csv'
```
cdmo, line 42:
```suggestion
```

These two files are produced programmatically by the rake task, so I don't think we need to make them a runtime configuration kinda thing - like the other config things can and do change in different environments and based on what we are trying to do with the traject execution. So, you could just delete them and then in `hash_to_hath` do:

```rb
#custom.rb line 217
     hathi_overlap_csv = "#{settings['hathi_overlap_path']}final_hathi_#{ht_format}_overlap.csv"
```

What do you think?

lib/traject/macros/custom.rb
cdmo: ok I had it wrong, it should have been "create a new _hash_  of arrays that is grouped by (i.e. keyed by) the first value of each array (the oclc number) not including the oclc number in the array" Pretty cool!

lib/traject/macros/custom.rb
```diff
@@ -205,12 +205,22 @@ def self.includes_oclc_indicators?(sf_a)
...
+        hathi_overlap_csv = "#{settings['hathi_overlap_path']}#{settings["hathi_#{ht_format}_overlap_file"]}"
+        CSV.read(hathi_overlap_csv)
+           .group_by(&:shift)
+           .each do |_oclc, ht_data|
+          ht_key = ht_format == 'mono' ? :ht_id : :ht_bib_key
+          ht_data.map! { |data| [ht_key, :access].zip(data).to_h }
+          ht_data.reject! { |data| data[:access] == 'deny' } if ht_data.length > 1
```
cdmo, line 223: I'm curious, is `ht_data.length` ever _more_ than 2? Not asking for a change to be clear, just curious. ht_data should be something like `{mono: 10, access: "deny"}`. And for the cases where there's only `1` that's because `access` is `nil` right?

spec/lib/traject/macros/custom_spec.rb
```diff
@@ -192,22 +192,58 @@
...
+
+    context 'for hathi records with mono item type' do
+      let(:ht_format) { 'multi' }
+
+      it 'produces hathi data correctly' do
+        expect(result['100000391']).to match [{ ht_bib_key: '012292266', access: 'allow' }]
+      end
+
+      it 'prefers the record with allow access if there are multiple copies with same oclc' do
+        expect(result['1000061']).to match [{ ht_bib_key: '005893467', access: 'allow' }]
```
cdmo, line 246: This is useful thanks!

lib/traject/macros/custom.rb
banukutlu: right ☝️

lib/traject/macros/custom.rb
banukutlu: the final filtered files  produced from the rake task should only have one of each allow or deny, because we are doing uniq on them.

cdmo: ok. cool. I should have just checked there. Thanks!

lib/traject/psulib_config.rb
banukutlu: this makes sense, I will try with your suggestion.

banukutlu: changes pushed

PR #222, opened 2020-07-17
fix for hathi data multiple deny copies bug

banukutlu:
Issue #224


cdmo:
Looking at the code and the issue, I don't see what the root cause of the problem is (the problem being that a record should be showing an ETAS link and it isn't). Is it related to https://github.com/psu-libraries/psulib_traject/pull/215#discussion_r444628466? Like not checking to see that there may be an `allow` when there are multiple `deny`s in a given row? Is this issue linked https://github.com/psu-libraries/psulib_blacklight/issues/601 ?

banukutlu:
> Looking at the code and the issue, I don't see what the root cause of the problem is (the problem being that a record should be showing an ETAS link and it isn't). Is it related to [#215 (comment)](https://github.com/psu-libraries/psulib_traject/pull/215#discussion_r444628466)? Like not checking to see that there may be an `allow` when there are multiple `deny`s in a given row? Is this issue linked [psu-libraries/psulib_blacklight#601](https://github.com/psu-libraries/psulib_blacklight/issues/601) ?

Yes that is related to that comment link, I was assuming uniq would give us only one allow or deny but that was wrong because uniq there is for the whole row not for the access column, there is the id or bib_key so we still ended up with multiples. This particular case is when there is no allow copy but multiple deny copies such as:

```
155131850,mdp.39076002651854,deny
155131850,mdp.39076002771645,deny
155131850,uc1.32106019703807,deny
```

I added specs for this exact case as well with the PR.

If there is an allow copy, allow is prefered  so that is another case, which would have been working OK (there is already a spec for that case.)

cdmo:
Right I know it's not a psulib_blacklight issue, of course.

So in a case where you have a record like 

```
155131850,mdp.39076002651854,deny
155131850,mdp.39076002771645,deny
155131850,uc1.32106019703807,deny
```

before this code change, what would the outcome be in the catalog? No ETAS link and nothing in the `hathidata_struct`?

banukutlu:
> Right I know it's not a psulib_blacklight issue, of course.
> 
> So in a case where you have a record like
> 
> ```
> 155131850,mdp.39076002651854,deny
> 155131850,mdp.39076002771645,deny
> 155131850,uc1.32106019703807,deny
> ```
> 
> before this code change, what would the outcome be in the catalog? No ETAS link and nothing in the `hathidata_struct`?

yes, no `hathidata_struct`

spec/lib/traject/macros/custom_spec.rb
```diff
@@ -233,9 +233,13 @@
...
         expect(result['1000021']).to match [{ ht_id: 'uc1.b3547182', access: 'allow' }]
       end
+
+      it 'selects only one deny copy if there are multiple copies with same oclc' do
+        expect(result['155131850']).to match [{ ht_id: 'mdp.39076002651854', access: 'deny' }]
+      end
     end
 
-    context 'for hathi records with mono item type' do
+    context 'for hathi records with multi/serial item type' do
```
cdmo, line 242: nice catch/fix

PR #225, opened 2020-07-22
Update psulib_config.rb

ruthtillman:
Removed 730 from uniform title for MARC reasons. Noticed while doing unrelated QA.


ruthtillman:
note: I also updated the documentation in the wiki.

Issue #230, opened 2020-08-05
Curbside Pickup change of logic for Hathi Public Domain


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Google-Books-and-HathiTrust-Integration-Links#curbside-pickup-change-of-logic-for-public-domain

eg. catkey: 737403
```
317082719    737403    multi    allow    pd
317082719    737403    multi    deny    ic
317082719    737403    multi    deny    und
```
 should be indexed with deny as ETAS record and I want It would be suppressed during ETAS period.

cdmo:
Just to clarify:

a `multi` with `allow` and `deny` should be indexed with the `ht_bib_key` and be marked as `true` for `etas` and the url to HathiTrust should be appended with the sign in suffix.

The two different parts are the sign in suffix and marking as etas.

PR #231, opened 2020-08-06
for hathi multi records with a mix of ETAS and PD, suppress allow

banukutlu:
Issue #230

### DIFF
--- a/lib/traject/macros/custom.rb
+++ b/lib/traject/macros/custom.rb
@@ -218,14 +218,18 @@ module Traject
         CSV.read(hathi_overlap_csv)
            .group_by(&:shift)
            .each do |_oclc, ht_data|
-          ht_key = ht_format == 'mono' ? :ht_id : :ht_bib_key
-          # For multi-volume records that include a mix of ETAS and public domain,
-          # prefer deny to suppress I want It button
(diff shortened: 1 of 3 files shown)


lib/traject/macros/custom.rb
```diff
@@ -218,12 +218,19 @@ def hathi_to_hash(ht_format)
...
            .each do |_oclc, ht_data|
-          ht_key = ht_format == 'mono' ? :ht_id : :ht_bib_key
-          ht_data.map! { |data| [ht_key, :access].zip(data).to_h }
+          ht_data.map! { |data| [ht_key(ht_format), :access].zip(data).to_h }
                  .uniq! { |data| data[:access] }
-          ht_data.reject! { |data| data[:access] == 'deny' } if ht_data.length > 1
+          ht_data.reject! { |data| data[:access] == ht_reject(ht_format) } if ht_data.length > 1
         end
       end
+
```
cdmo, line 226:
```suggestion
      private

```

Since only other methods in this mixin send these messages I think they should be private.

banukutlu: oh yes thanks! I thought about adding this then forget LOL!

PR #242, opened 2020-09-17
had to add config setup to hathi rake task

banukutlu:
there might be a better way to do this 🤔


cdmo (dismissed):
looks good to me, but needs a rubocop fix https://travis-ci.com/github/psu-libraries/psulib_traject/builds/184991408#L731

cdmo (approved):

PR #246, opened 2020-09-25
Adds open access option for access_facet, plus a bit of refactoring

cdmo:
#245

Issue #241, opened 2020-09-10
Index 099a so that it's in keyword search


ruthtillman:
Cataloging is making the 099a into a canonical field for mapping special collections accessioning numbers. It does not appear to be indexed in keyword search.

Examples:
* Search for 09715 should get https://catalog.libraries.psu.edu/catalog/16418026
* 09844 should get https://catalog.libraries.psu.edu/catalog/21330529 
* 09981 should get https://catalog.libraries.psu.edu/catalog/26782112

ruthtillman:
If this is something fairly easy to just get into the keyword index (no need to display), then can we add it to current indexing work? If it's more complex, we can wait a little bit on it.

cdmo:
@ruthtillman when I do a search 

* for `09715` https://catalog.libraries.psu.edu/catalog/16418026 shows as the number 2 (out of 2) results.
* for `09844` https://catalog.libraries.psu.edu/catalog/21330529 shows as the number 2 (out of 2) results.
* for `09981` https://catalog.libraries.psu.edu/catalog/26782112 shows as the number 6 (out of 6) results.

So the field is indeed indexed. 

Or is this request for the placement in the search results for these _special collections accessioning numbers_ to be higher? As in the above 3 records should all be _the first_ result in searches for the special collections accessioning number?

ruthtillman:
🤔 weird, I wasn't getting those when I tried it--I was just getting 1 result for those first two, the one with the 856 in it.

It would be helpful if it were weighed higher, yes. Especially since it's a numerical thing so that shouldn't really interfere with other searches.

cdmo:
👍 

Very odd it wasn't there before. I take that these accessioning number identifiers have been around a long time right?

ruthtillman:
totally. I don't know if I searched at a weird time...

cdmo:
🤔 

Side thought: We should make a relevancy testing suite some day just to make sure we're always good with what we expect. 

Anyway, I'm working one raising relevancy for this ID now 👍

cdmo:
@ruthtillman it looks like the `099a` might be used for purposes other than the special collections accessioning number? Like this govdoc: https://catalog.libraries.psu.edu/catalog/55106. This old newspaper (?) has a weird thing going on in the 099a too: https://catalog.libraries.psu.edu/catalog/126675

[...]

Issue #244, opened 2020-09-22
Index MARC fields 255, 775, and 793 for display


ruthtillman:
Over the last few months, users have identified 3 more MARC fields which should be displayed in records.

| field | label | documentation |
| --- | --- | --- |
| 255abcdefg | Cartographic Mathematical Data | https://github.com/psu-libraries/psulib_blacklight/wiki/Publication-Info-Fields-Index-and-Display |
| 775\|0*\|iabcdefghikmnor | Other Edition | https://github.com/psu-libraries/psulib_blacklight/wiki/Publication-Info-Fields-Index-and-Display |
| 793a | Collection | https://github.com/psu-libraries/psulib_blacklight/wiki/Title-Fields-Indexing-and-Display |

Their locations for display are here: https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields with temporary notes dated 2020-09-22 to make them easy to find, see also Blacklight issue:

PR #247, opened 2020-09-28
241 accessioning numbers

cdmo:
#241 

#246 will need to be reviewed first, this branch built on top of that.


spec/lib/traject/psulib_config_spec.rb
```diff
@@ -4,26 +4,37 @@
...
 
-  describe 'Record with music numeric should have semicolons for all but subfield e' do
-    let(:field) { 'music_numerical_ssm' }
-    let(:music_383) do
-      { '383' => { 'ind1' => '1', 'ind2' => '0', 'subfields' => [{ 'b' => 'op. 36' },
-                                                                 { 'b' => 'op. 86' },
-                                                                 { 'b' => 'op. 35' },
-                                                                 { 'e' => 'Bach' },
-                                                                 { 'e' => 'Motzart' }] } }
+  describe 'id' do
```
banukutlu, line 15: id and music numeric specs just re-ordered right?

cdmo: yeah that's what I meant in the commit message, sorry, not clear.

I also have to say, as a bigger topic, do we want the new spec I added for this? I think the answer is "yes" but we also have dozens of straightforward `extract_marc` fields un-tested.... This is relevant for work required in https://github.com/psu-libraries/psulib_traject/issues/244 too.  Again my leaning is that "yes" they need tests just for clarity. But I dont know that I want to add tests for all of our fields. How much value do we get for these tests?

banukutlu: I think we initially decided not to add specs for just straightforward extract_marc fields, and just add ones for that involves additional logic. 
Saying that, I think it is OK even nice to have a spec for a newly added field, kinda gives a piece of mind. 

If we wanna consider about adding a spec for those other fields too, I agree that might be a work that might not hold as much value in the end but maybe we can add several fixtures that captures all the fields and somehow find a way to just test result of the whole import, kinda like a feature specish.

banukutlu (approved):
👍

spec/lib/traject/psulib_config_spec.rb
cdmo: think we're on the same page 👍

Issue #186, opened 2019-08-01
Address govt publication complication w/university presses


ruthtillman:
When "University" and "Press" aren't together, in the press's name, it still shows up as a government document. "Penn State University Press" would be caught. "University of Michigan Press" would not. Can we change this:

https://github.com/psu-libraries/psulib_traject/blob/80d4014d1f43e39dd1ba0ae917fb8cc790a0ae38/lib/marc_format_processor.rb#L120-L124

to a contains search vs. a phrase search?

Example of what university press names tend to look like: https://catalog.libraries.psu.edu/catalog?utf8=%E2%9C%93&op=AND&all_fields=&title=&author=&subject=&identifiers=&series=&publisher=university+press&f_inclusive%5Bformat%5D%5B%5D=Government+Document&range%5Bpub_date_itsi%5D%5Bbegin%5D=&range%5Bpub_date_itsi%5D%5Bend%5D=&sort=score+desc%2C+pub_date_itsi+desc%2C+title_sort+asc&search_field=advanced&commit=Search

banukutlu:
@ruthtillman is this closed mistakenly, maybe?

ruthtillman:
Yes, I was closing the others and accidentally hit it too. Thanks.

ruthtillman:
Updated the wiki to clarify that the goal of this change is to find every instance of "University" AND "Press" in 264s.

University of Michigan Press and Penn State University Press should both be caught.

banukutlu:
@ruthtillman do we need to consider any non-english representations?

ruthtillman:
Good question -- we do not because we should have English representations of everything, the others will be duplicates.

ruthtillman:
Looks good! Couple edge cases are slipping through but that's a data issue re: catalogers not always including "Press" which isn't a blocker. Huge improvement comparing to current catalog. Thank you!

ruthtillman:
Update -- after more review, I realized we should just limit the test to "University." This will catch some edge cases and while still being sufficient for our needs.

ruthtillman:
this is good!

PR #248, opened 2020-09-29
format processing update for university press 

banukutlu:
Issue #186


lib/marc_format_processor.rb
```diff
@@ -120,7 +120,7 @@ def government_document?(record)
   # Check if 260b OR 264b contain variations of "University Press"
   def university_press?(record)
     field_260b_264b = Traject::MarcExtractor.cached('260b:264b', separator: nil).extract(record)
-    field_260b_264b.grep(/University Press/i).any?
+    field_260b_264b.grep(/University|Press/i).any?
```
cdmo, line 123: This is saying match anything that has University OR Press. Isn't that too permissive? The following would be matches:

```
University Press
University Very Special Press
Press
Pressers
University Machine
University Press
University Very Special Press
Press
Pressers
University Machine
Impressive Stuff
```

I think it'd be better to do `/University\b.*\bpress\b/i`, see https://rubular.com/r/YlKC4rvzpQwRPk

banukutlu: good point! will change, thanks

cdmo (approved):
👍

Issue #239, opened 2020-09-01
Update format index to add Articles


ruthtillman:
Update formats mapping so that if leader[6] == `a` AND leader[7] == `a` the format is Article

https://github.com/psu-libraries/psulib_blacklight/wiki/Formats#formats-mapping

Sample catkeys:

* 31199600
* 31199627
* 31201441

PR #249, opened 2020-09-29
adding format article

banukutlu:
#239


lib/marc_format_processor.rb
```diff
@@ -45,32 +45,40 @@ def resolve_leader(record)
```
cdmo, line 45: I think I'd like to see this method renamed `resolved_leader_byte_6` (is now `resolve_leader`)

I also kinda wonder how it would look to give these methods more human names and hide their machine-code complexity? Like instead of `resolved_leader_byte_6` it could be called `type_of_record` and then `resolve_leader_byte_7 ` could be `bibliographic_level`. Cribbing from https://www.loc.gov/marc/bibliographic/bdleader.html

I wonder if that'd be easier to read and understand. I kinda like the idea of using method names to help label what the marc data is trying to discern about the record..

banukutlu: sounds good to me, changing them

cdmo: I do really like this better 👍 Thanks

one small thing:

`holdings_formats`... kinda not getting this one sorry. `949t` seems like "special local possibilities for format." Is that right? Like these are specially defined by penn state formats. Maybe `local_formats` might be better? Or something?

cdmo (changes requested):
see note in comment https://github.com/psu-libraries/psulib_traject/pull/249/files#r497894855

cdmo (approved):

PR #255, opened 2020-10-02
Need to double quote the strings with spaces for Splunk to be happy

cdmo:


lib/tasks/incrementals.rake
```diff
@@ -8,29 +8,32 @@ namespace :incrementals do
   desc 'Adds to the index'
   task :import, [:period] do |_task, args|
     indexer = Traject::Indexer::MarcIndexer.new
-    indexer.logger.info "name=Sirsi Incremental message=Indexing operation beginning task=#{args[:period]} import progress=start"
+    indexer.logger.info 'name="Sirsi Incremental" '\
+                        'message="Indexing operation beginning" '\
+                        "task=\"#{args[:period]} import\" "
+                        'progress=start'
```
banukutlu, line 14: progress does not need double quotes?

cdmo: nope
:)

cdmo: oh hold on

cdmo: needed a trailing backslash on line 13

lib/tasks/incrementals.rake
```diff
@@ -8,29 +8,32 @@ namespace :incrementals do
...
     end
 
-    indexer.logger.info 'name=Sirsi Incremental '\
-                        "message=Indexing #{target} "\
-                        "task=#{args[:period]} import "\
-                        'progress=in progress'
+    indexer.logger.info 'name="Sirsi Incremental" '\
+                        "message=\"Indexing #{target}\" "\
+                        "task=\"#{args[:period]} import\" "\
+                        'progress="in progress"'
```
banukutlu, line 28: no need for double quotes here?

cdmo: line 28?

banukutlu: yup there is double quotes here for `in progress` but for example not on line 14 for `start`

cdmo: yes because it's two words. splunk interprets spaces as the end of the key/value pair.

I think this PR will make splunk happier, but I'm not 100% sure. Hard to test in dev. I could deploy this branch and see how it does?

banukutlu (approved):
👍

PR #262, opened 2020-10-18
Swaps out reporting document ids with document count

cdmo:


lib/tasks/incrementals.rake
```diff
@@ -85,9 +85,9 @@ namespace :incrementals do
     end
 
     if ids.any?
-      deleted_ids = ids.join ','
+      formatted_count = ids.count.to_s.chars.reverse.each_slice(3).map(&:join).join(',').reverse
```
banukutlu, line 88: 👀

banukutlu (approved):

lib/tasks/incrementals.rake
banukutlu: just to note here: this is to format `8888888` as `8,888,888`

cdmo: could add a quick comment. I was actually kinda surprised that `Integer` didn't have a format option 🤷

banukutlu: yup, same here

Issue #121, opened 2019-04-10
Location mapping


cdmo:
https://github.com/psu-libraries/psulib_blacklight/wiki/Library-Faceting-and-Locations-Management

Required by https://github.com/psu-libraries/psulib_blacklight/issues/269

ruthtillman:
Controlled list of excludes is here: https://github.com/psu-libraries/psulib_blacklight/wiki/Library-Faceting-and-Locations-Management#locations-advanced-search

ruthtillman:
👍

PR #264, opened 2020-11-23
locations mapping

banukutlu:
#121


.rubocop_todo.yml
```diff
@@ -57,7 +57,7 @@ Style/MixinUsage:
   Exclude:
     - 'lib/traject/psulib_config.rb'
 
-# Offense count: 57
+# Offense count: 59
```
cdmo, line 60: might want to add an exclude for `setting.yml` or turn that large array into a format like 

```yml
location_excludes:
  - item 1
  - item 2
...
```

small thing

.rubocop_todo.yml
```diff
@@ -35,7 +35,7 @@ Metrics/MethodLength:
 # Offense count: 1
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ModuleLength:
-  Max: 174
+  Max: 180
```
cdmo, line 38: Wonder if we shouldn't add an exclude here for `psulib_config.rb` 🤔

spec/lib/traject/macros/custom_spec.rb
```diff
@@ -263,4 +263,15 @@
       end
     end
   end
+
+  describe '#exclude_locations' do
+    let(:location_1) { { '949' => { 'ind1' => '', 'ind2' => '', 'subfields' => [{ 'l' => 'DOCUSMF-DN' }] } } }
+    let(:location_2) { { '949' => { 'ind1' => '', 'ind2' => '', 'subfields' => [{ 'l' => 'RESERVE-HN' }] } } }
+    let(:location_3) { { '949' => { 'ind1' => '', 'ind2' => '', 'subfields' => [{ 'l' => 'AVAIL_SOON' }] } } }
+    let(:result) { @indexer.map_record(MARC::Record.new_from_hash('fields' => [location_1, location_2, location_3], 'leader' => leader)) }
```
cdmo, line 271: latest version of rubocop seem to wanting to use let memoization less. 🤷‍♂️ It does feel easier to read and it is shorter.

```suggestion
    location_1 = { '949' => { 'ind1' => '', 'ind2' => '', 'subfields' => [{ 'l' => 'DOCUSMF-DN' }] } }
    location_2 = { '949' => { 'ind1' => '', 'ind2' => '', 'subfields' => [{ 'l' => 'RESERVE-HN' }] } }
    location_3 = { '949' => { 'ind1' => '', 'ind2' => '', 'subfields' => [{ 'l' => 'AVAIL_SOON' }] } }
    result = @indexer.map_record(MARC::Record.new_from_hash('fields' => [location_1, location_2, location_3], 'leader' => leader))
```

lib/translation_maps/locations.properties
```diff
@@ -0,0 +1,709 @@
+ACAD-BLDG = Submit request for annexed material
```
cdmo, line 1: We can't get this from a web service call, right?

lib/traject/macros/custom.rb
```diff
@@ -226,6 +226,13 @@ def hathi_to_hash(ht_format)
         end
       end
 
+      def exclude_locations
+        lambda do |_record, accumulator|
+          accumulator.reject! { |value| ConfigSettings.location_excludes.include?(value) }
+          accumulator.reject!(&:empty?)
```
cdmo, line 232: is this the same as `accumulator.compact`? Just eliminating empty items?

cdmo (changes requested):
Couple of question/thoughts, looks good though.

lib/traject/macros/custom.rb
banukutlu: ah yes most likely, I just forget about that method :) I will check and change it so it is consistent with the rest

cdmo: I do love me a pretzel colon operator - but yeah I think if it's the same as compact that probably makes more sense (being nitpicky I know sorry)

.rubocop_todo.yml
banukutlu: 180 is from `lib/traject/macros/custom.rb`

lib/translation_maps/locations.properties
banukutlu: I am not sure honestly, can check with Maryam

.rubocop_todo.yml
cdmo: oh right!

lib/translation_maps/locations.properties
cdmo: ok, if she says yes we can follow up with a new PR to replace 👍

cdmo (approved):

lib/translation_maps/locations.properties
banukutlu: exactly what I was thinking 👍 :)

PR #265, opened 2020-11-24
Replaces HathiTrust struct with a simple allow/deny access string including a fix to ensure proper handling of resolving access level

cdmo:
Can you give this a once over? I am still yet to remove the unneeded specs and make new ones, but I want to see if the new to_field definition makes sense to you.


lib/traject/psulib_config.rb
```diff
@@ -52,14 +51,13 @@
...
-# HATHI_MULTI_OVERLAP constant is expected to be a three column csv of oclc number, ht_bib_key and access info.
-# These are generated by running `rake hathitrust:process_hathi_etas`.
-Traject::Macros::Custom::HATHI_MONO_OVERLAP = hathi_to_hash('mono')
-Traject::Macros::Custom::HATHI_MULTI_OVERLAP = hathi_to_hash('multi')
-
 ATOZ = ('a'..'z').to_a.join('')
 ATOU = ('a'..'u').to_a.join('')
+HATHI_OVERLAP_REPORT = CSV.read(ConfigSettings.hathi_overlap_path, headers: true, col_sep: "\t")
+                           .reject { |hr| hr['access'].nil? }
+                           .reject { |hr| ['2168941'].include?(hr['local_id']) }
```
cdmo, line 58: this line is about the excludes list - I'll store this in a constant for the next iteration

lib/traject/psulib_config.rb
```diff
@@ -208,8 +206,28 @@
...
+
+  # Is the catkey in the Hathi overlap report? If so, write down allow or deny.
+  # relevant is an array of finds (often just one)
+  relevant = HATHI_OVERLAP_REPORT.find_all { |hr| hr['local_id'] == catkey  }
+
+  accumulator << if relevant.any?
+    # When there's more than one
+    if relevant.count > 1
+      access_array = relevant.collect {|hr| hr['access']}
+      if access_array.include? 'deny'
```
cdmo, line 221: I don't know that this is right but I think it is. Basically if there are dupe items in the overlap then I'm saying "deny" wins. 🤔 Should it be the other way around? Does mono versus multi in this case make a difference?

cdmo: in cases like 

```
2384942	1099502	mono	deny	ic
2384942	1099502	mono	allow	pd
2384942	1099502	multi	deny	ic
2384942	1099502	multi	allow	pd
```

`allow` wins (as per slack). I'll update the wiki.

cdmo:
I feel like the CC issue to fix is not important, or rather I'm not sure how I could make it less complex.

A PR in blackcat to make use of this will come soon, working on it now

config/settings/test.yml
```diff
@@ -1,2 +1,2 @@
 processing_thread_pool: 5
-hathi_overlap_path: spec/fixtures/hathitrust/
+hathi_overlap_path: spec/fixtures/hathitrust/overlap.tsv
```
banukutlu, line 2: very minor, not important but renaming it as to sth like `hathi_overlap_file` would it make more sense?

cdmo: isn't that implied by the directory it is in?

spec/lib/traject/access_facet_spec.rb
```diff
@@ -52,36 +52,10 @@
...
-      end
-
-      it 'empty when record has a hathitrust copy with deny access' do
-        result = @indexer.map_record(MARC::Reader.new(File.join(fixture_path, 'access_hathi_deny.mrc')).to_a.first)
-        expect(result['access_facet']).to be_nil
-      end
+    it 'produces Online when a record has a hathitrust copy of any kind' do
+      id = { '001' => '1000065' }
+      result = @indexer.map_record(MARC::Record.new_from_hash('fields' => [id], 'leader' => '1234567890'))
+      expect(result['access_facet']).to contain_exactly 'Online'
```
banukutlu, line 58: @cdmo would a deny record post-etas have online?

spec/lib/traject/psulib_config_spec.rb
```diff
@@ -94,4 +95,27 @@
...
+      expect(result['ht_access_ss']).to contain_exactly 'allow'
+    end
+
+    it 'returns "allow" when duplicates are both mono and multi/serial and the access level is mixed' do
+      id = { '001' => '1099502' }
+      result = @indexer.map_record(MARC::Record.new_from_hash('fields' => [id], 'leader' => leader))
+
+      expect(result['ht_access_ss']).to contain_exactly 'allow'
+    end
+  end
```
banukutlu, line 120: we do not need any specs for deny case?

spec/lib/traject/access_facet_spec.rb
cdmo: No it wouldn't, I'll fix this.

spec/lib/traject/psulib_config_spec.rb
cdmo: I'll add some.

cdmo: Counting all possible combinations of duplicate, no duplicates, allows, denies and mixes there are around 20 possible scenarios I think. For example here are the possible options for just mono (not multis, not mono/multi mixes):

1. Mono, no dupes: allow
1. Mono, no dupes: deny
1. Mono, no dupes: blank
1. Mono, dupes: deny
1. Mono, dupes: allow
1. Mono, dupes: blank
1. Mono, dupes: mix

Thoughts? Just trying to figure out which should be covered.

cdmo: I think I'll just write notes for "similar scenarios" on things like "mono" with one access level because it's the same as multi with one access or dupes that are all the same level of access 🤔

cdmo: should be in better shape now. I think it was definitely a problem that a dupe multi with mix access wasn't covered. thanks

spec/lib/traject/access_facet_spec.rb
cdmo: covered now, thanks for catching this!

banukutlu: great, thanks!

banukutlu (approved):
looks good to me 👍

Issue #263, opened 2020-11-03
Update link indexing to include two blank indicators on 856


ruthtillman:
While not ideal as it provides very little info about the resource linked, the double null indicator for `856|  |` is a valid use case. Because some of the records we're getting from vendors use it, we'll need to add it to our indexing. Treat as a full-text link.

banukutlu:
@ruthtillman do you have an example catkey?

ruthtillman:
Unfortunately, we fixed the catkey where it was reported. @mkutch is this something you could get via a query?

mkutch:
@ruthtillman I'll look... in the meantime approx when was the catkey changed.... I want to see if it still exists in DEV (which hasn't been synched for months) to see what it looks like... So there is no $u subfield?

ruthtillman:
@mkutch catkey is: 6946582 and was only fixed at the end of Oct. Two blank indicators -- there's still a $u.

mkutch:
catkeys
`1174303`
`23786738`
`2206901`
`2187375`

do you need more?

banukutlu:
> catkeys
> `1174303`
> `23786738`
> `2206901`
> `2187375`
> 
> do you need more?

thanks these are good 👍

banukutlu:
@ruthtillman could you please update https://github.com/psu-libraries/psulib_blacklight/wiki/URL-Fields as well?

ruthtillman:
Done!

ruthtillman:
These look great! Good to deploy.

PR #266, opened 2020-11-30
include 856u with two blank indicators to fulltext links

banukutlu:
#263


lib/traject/macros/custom.rb
```diff
@@ -132,7 +132,7 @@ def sought_link_data_exists?(link_type, field)
       end
 
       def fulltext_link_available?(ind1, ind2, url_label)
-        (ind2 == '0' || ind1 == '4') && !NOT_FULLTEXT.match?(url_label)
+        (ind2 == '0' || ind1 == '4' || (ind1.strip.empty? && ind2.strip.empty?)) && !NOT_FULLTEXT.match?(url_label)
```
cdmo, line 135: I'm not sure I see why `strip` is needed. Looking at the spec it looks like an empty string is passed... 🤔

cdmo (changes requested):
One question, other than that a nice tidy PR 👍

lib/traject/macros/custom.rb
banukutlu: yeah this was interesting to me too, the example record I was testing came with whitespace in its indicators like `" "` so I had to add that strip 🤔

cdmo: ah ok, i was just going by the spec. Did you also see cases where the string was just a true empty string (i.e., `""`)?

banukutlu: I tried the example records we have in the issue and they all had whitespace in both indicators. of course I am not sure if that is the case always.

```
1174303
23786738
2206901
2187375
```

cdmo: Ah I was just thrown off by the spec which passes empty string not `" "`

cdmo:
Strange that the build failed in a new way. Did you see that? https://travis-ci.com/github/psu-libraries/psulib_traject/builds/205495909#L794

lib/traject/macros/custom.rb
banukutlu: yeah, I will make the specs like this `" "`

cdmo: 🙏 danke

cdmo (changes requested):
just needs a rebase

cdmo (approved):

PR #273, opened 2021-01-22
Adds rake task for getting last incremented solr collection

cdmo:


lib/psulib_traject/solr_manager.rb
```diff
@@ -0,0 +1,59 @@
...
+    def connection
+      @connection ||= Faraday.new(ConfigSettings.solr.url) do |faraday|
+        faraday.request :multipart
+        faraday.adapter :net_http
+      end
+    end
+  end
+
+  # Lightweight Solr collection abstraction
+  class SolrCollection
```
banukutlu, line 44: nice 👍 maybe we implement this abstraction to BlackCat solrmanager?

banukutlu (approved):
👍

lib/psulib_traject/solr_manager.rb
cdmo: Yeah, I thought this was an improvement too, wanted to hear your thoughts. I think I wish SolrManager was just a separate plugin we could use anywhere, if time were infinite that's what I'd prefer

banukutlu: yes! how nice that would be 💯

PR #283, opened 2021-04-09
Use standard as our Ruby linter

awead:
Installing Niftany, our current linter, posed some problems because of its dependency on erb-linter. Since this probject isn't Rails, using Niftany will not work. As an alternative, we can use the standard gem.

The rules differ from our Niftany gem, but the advantage is there is no need for shared Rubocop rules. Simply install the gem and run the task.

## Why Are We Doing This?

I think it might make life easier, if we're willing to let go of some styling decisions.

In order to share linter definitions in different projects, we've used Niftany, which wraps a bunch of different linters into one gem: Rubobop for Ruby (including RSpec), erb-lint for erb files, and sass-lint for css. The problem with this is that this means a lot of dependencies. For a project like traject, there's _only_ Ruby, so we don't need all the other gems, and in fact, the project won't build at all if you try to add the Niftany gem!

## Why Can't We Just Keep Using Rubocop?

We can. This is a proposal to switch to standard Ruby instead of Rubocop. You can read about it here:

https://github.com/testdouble/standard

The idea is that our Ruby style is dictated by the gem. We don't have to decide how we want to do things, and maintain separate Rubocop definitions, we just use the gem.

## Will This Replace Niftany?

Probably not. We can still keep using Niftany for Rails projects, and add standard Ruby to Niftany instead. We'd use Niftany for any Rails project, and the standard gem for any plain Ruby code. The advantage is they'd both be sharing the same style definitions via the standard gem.

## Can We Not Do This?

Absolutely. We don't have to to this. We don't use the the gem and continue to use our custom Rubocop definitions via Niftany. But, if we want to include style definitions in a plain Ruby project like this, we'd need to copy them from Niftany or just maintain them separately. If style changes were made to Niftany, they wouldn't make it over here unless we added them manually.

## Are There Alternatives?

Sure... we could make another gem that's just the Ruby/Rubocop definitions, but that's more maintenance. Again, I'm looking at ways to make things easier (?) and if we let go of some style decisions, we could just use Standard Ruby and move on. But if we don't want to do that, I'm 💯 fine with it and we can keep things the way they are.


banukutlu:
@awead I am still torn on this one, I agree that not having to make style decisions and let the standard gem dictate is making our life easier but still double quotes making me 😭  Do you have any non-rails just ruby projects? wondered what do you use for those?

awead:
@banukutlu I can see if there’s a way to repurpose Niftany for this.

banukutlu:
> @banukutlu I can see if there’s a way to repurpose Niftany for this.

do not worry if it looks like it will require work/time, but it would be awesome if it can be! thanks Adam!

awead:
Closing in favor of #297

PR #290, opened 2021-06-02
Fix microform classification

### LINKED ISSUE #287: Fix Microform Classification
From E.N. email:

```
We started by using the new catalog.  We got 8,661 results for: 
Campus: University Park  
Format: Government Document 
Media Type:  Microfilm/Microfiche 

A search in the legacy CAT brings up 71,340 records for
Material Type: Microfilm/Microfiche
Location: Paterno - 2nd Floor - U.S. Documents
```

(note: the above numbers are for a specific search, we have more total microforms than just 71,340 but after we complete the ticket we should match them in an Advanced search for parameters: `Format: Microfilm/Microfiche` & `Location: Paterno - 2nd Floor - U.S. Docs` - run in CAT as well just to be sure numbers haven't changed in interim.)

Issue identified -- our media format issue relies on MARC fields which are not necessarily present if the Microformat record is a reused print record which has not been properly adapted into a Microformat record. This means that our MARC-based indexing will identify it as a print volume vs. a microformat. This should not affect things related to bound-in. It's actually easy to identify all Microformats using the following criteria:

IF the record includes at least on 949t which has the value MICROFORM, it's of that type.

Because our default use of 949t goes to formats, we can also use this moment to change Microformats from Material Type to Format per request of microformats library (while not strictly ontologically correct, it's something the users who do this search most often have requested and, other than ontological considerations, there's no reason not to).

Steps:

* [x] Ruth update wiki
* [ ] remove MARC-based indexing for Microformats as Material Type
* [ ] add 949t-based indexing for Microformats as Format
* [ ] add format icon for display: https://github.com/psu-libraries/psulib_blacklight/issues/749
* [ ] review spec

sample catkeys: `1477303, 1939940, 1939929`.

Key files:

- MARC based indexing is in: these two lines https://github.com/psu-libraries/psulib_traject/blob/main/lib/marc_media_type_processor.rb#L47-L48 (we need to retain the others in that section) and this line https://github.com/psu-libraries/psulib_traject/blob/main/lib/marc_media_type_processor.rb#L152
  - note: these were supposed to catch properly coded 007 fields and then the 300a was supposed to be a backup for records which didn't have good 007 but at least had been turned into microforms records, which we've learned many have not.
- where we set up the 949t indexing https://github.com/psu-libraries/psulib_traject/blob/main/lib/translation_maps/formats_949t.properties just need to add `MICROFORM = Microfilm/Microfiche`

awead:
Refactors MarcMediaTypeProcessor and removes its microfilm format logic in favor of using the translation map file.

Fixes #287

## Issues

Making the change to `formats_949t.properties` required changing the tests, which now seem to overly favor microfilm as a format. The provided marc fixture files seem to support this: they all look like actual microfilms. We might want to provide additional tests, with supporting fixtures, to ensure that the other formats are getting tested properly.


ruthtillman (commented):
I'll leave approval to Banu but from the process perspective of assigning media and format types, things look good. Happy to review once we've run an index.

awead:
This line jumped out at me: https://github.com/psu-libraries/psulib_traject/pull/290/files#diff-6fb6123bac6081673ad76669b5d71b46669f036707072682853683e2c0358c64R105

It seems to suggest that Microfilm is the "other" of the formats. It seems like we ought to have an example marc record that is truly `Other` so we have something to test against.

ruthtillman:
@awead while in their **current** status, (e.g. at Bindery etc) these two items are Other. https://catalog.libraries.psu.edu/?f%5Baccess_facet%5D%5B%5D=Other

so if we wanted to get a MARC record example for "Other" to put in our tests, we could download either of them.

lib/marc_media_type_processor.rb
```diff
@@ -1,114 +1,64 @@
 # frozen_string_literal: true
 
-# A tool for classifying MARC records using a combination of data from the
-# record's leader and some 949ts to assign media types to records
+# A tool for classifying MARC records using a combination of data from the record's leader and some 949ts to assign
+# media types to records
 class MarcMediaTypeProcessor
-  def initialize
-    freeze
```
banukutlu, line 7: @awead fwiw we had seen concurrency issues so ended up addind freeze to make it stateless and thread-safe, related PR: https://github.com/psu-libraries/psulib_traject/pull/122/files. So, I just wondered about that you haven't seen such behavior?

awead: Interesting! Yes, you would run into thread issues with JRuby. I think we can make it thread-safe and refactor at the same time. Let me update it.

banukutlu:
@awead Looks great to me! I agree, as you mentioned we need additional specs or just need to update to use other records for the current types that are now resolving as microfilm.

banukutlu:
> @ruthtillman @banukutlu I don't think we have to have additional marc files to test each of the different scenarios, we just have to test the logic. I kind of hinted at that here: https://github.com/psu-libraries/psulib_traject/pull/290/files#diff-6fb6123bac6081673ad76669b5d71b46669f036707072682853683e2c0358c64R107
> 
> You can fake a record with the requisite parts, or just mock it completely. I'll work something up and post it.

yup that works too 👍  another similar approach we used in other specs https://github.com/psu-libraries/psulib_traject/blob/main/spec/lib/traject/subjects_spec.rb#L15. Either way is fine, with fixture files specs look a bit less crowded but it does not matter which everway you like is fine 👌

banukutlu:
> @ruthtillman @banukutlu I don't think we have to have additional marc files to test each of the different scenarios, we just have to test the logic. I kind of hinted at that here: https://github.com/psu-libraries/psulib_traject/pull/290/files#diff-6fb6123bac6081673ad76669b5d71b46669f036707072682853683e2c0358c64R107
> 
> You can fake a record with the requisite parts, or just mock it completely. I'll work something up and post it.

oh yeah cannot test every different scenario/format etc, did not do that before either, as you said just testing the different logic makes sense..

awead:
@banukutlu Ok, I've added two more commits to refactor MarcFormatProcessor. I've removed some of the duplicated tests and added additional ones should should get the test coverage up to %100 for the processor.

spec/spec_helper.rb
```diff
@@ -32,7 +32,18 @@
 # it.
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
+
+def indexer
+  @indexer
+end
+
 RSpec.configure do |config|
+  config.before(:all) do
```
banukutlu, line 41: this refactor is :tada: 👌  🎉   neat and cleaner!

banukutlu (commented):
thanks @awead!

banukutlu (approved):

PR #293, opened 2021-06-08
Update Docker and Solr related tasks

awead:
Changes the Docker rake tasks to use "psul_catalog" as the collection name, which matches the name in settings.yml. The solr tasks file is renamed to "solr" to lineup with the namespace.


banukutlu (approved):
👌

PR #297, opened 2021-06-10
Use Niftany Ruby and RSpec rules

awead:
We can't use the Niftany gem directly, but we can use the rules by importing them.


banukutlu (approved):
thanks @adam! 👌

PR #299, opened 2021-06-10
Reorganize repository code

awead:
The files in the lib and spec directories are shuffled around and renamed. Now, the application can be loaded with a single `require 'psulib_traject'` statement and the structure of the spec directory maps to the structure of the lib directory. This makes it easier when to switch from a class to its given unit test. A lot of editors can find them automatically that way.

Tests that have no associated class are moved to spec/integration because they're testing the entire system instead of a specific class. This is like the unit versus "feature" test distinction in Rails.

Lastly, the config file that runs the whole show is moved to `config/` as a means of keeping the application code separate from the executable portion of the code.


banukutlu:
@awead When I run traject on local, I am getting the below error on this branch. Have you gotten this one?

```
➜  psulib_traject git:(reorg) ✗ bundle exec traject -c config/traject.rb solr/sample_data/1055577.marc
...
bundler: failed to load command: traject (/Users/bzk60/.rbenv/versions/jruby-9.2.11.1/bin/traject)
NameError: uninitialized constant Traject::MarcCombiningReader
 org/jruby/RubyModule.java:3760:in `const_missing'
...
```

awead:
@banukutlu Sorry, missed one. I should be fixed now.

banukutlu:
Requires psu-libraries/psulib_blacklight_deploy#111

banukutlu (approved):
lgtm 👍  as mentioned this change requires psu-libraries/psulib_blacklight_deploy#111

PR #304, opened 2021-06-16
Configure Solr's port

awead:
This allows us to run Solr on a different port, which is helpful if you have multiple apps running with their own Solr, or want to test different versions on different ports.

This will likely all go away when we have a proper Docker compose setup, but this is mostly for me so I can run Scholarsphere and the catalog simultaneously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,66 +7,83 @@
 
 ### Java
 To run JRuby you will need a JRE (the JVM runtime environment) version 7 or higher.
-```
-$ java --version
-  java 9
-  Java(TM) SE Runtime Environment (build 9+181)
-  Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)
-```
+
+    $ java --version
+      java 9
+      Java(TM) SE Runtime Environment (build 9+181)
+      Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)
 
 ### Ruby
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -7,66 +7,77 @@
...
 
-    If Docker Solr isn't running (check `docker ps`) run
+When using jruby, traject will use multiple threads, but we want to tailor that to our system. In
+`config/settings.local.yml` add:
 
-    ```
-    $ bundle exec rake solr:up
-    ```
+    hathi_overlap_path: spec/fixtures/hathitrust/overlap.tsv
+    processing_thread_pool: 5
```
banukutlu, line 55: thanks for adding this too @awead 👌

README.md
```diff
@@ -7,66 +7,77 @@
...
+   
+## Build an Index
+
+Start Solr via the Docker container
+    
+    $ bundle exec rake docker:first_start
```
banukutlu, line 61: @awead should this be `bundle exec rake solr:up` maybe 🤔

banukutlu: Also, I could get `docker:up` to work by only switching back to backticks rather than `exec`, wondering if I am missing sth, does `docker:up` works for you?

awead: @banukutlu try it again. I've simplified it a bit.

banukutlu (approved):
👌

PR #306, opened 2021-06-22
Add additional formats when microfilm already exists

### LINKED ISSUE #303: Update Microfilm indexing to catch other formats
Microfilm record indexing currently stops after indexing the one kind of thing but we need to record multiple formats for a record if there's also serials/periodicals on it: https://catalog.libraries.psu.edu/catalog/483004

Essentially wherever it falls in things it should test for microfilms AND test for any other formats which might match.

We're likely to see the following combos: Microfilm/book, Microfilm/serial, Microfilm/archival material.

awead:
This does some additional refactoring of the existing `Processors::Format` class and adds two additional classes, `PreferredFormat` and `RecordType`.  The former serves as an "early exit" strategy that will return the appropriate formats if they're found first. If no preferred formats exist, then the latter will add additional formats if the existing formats include microfilm.

Note: I've removed the "Instructional Material" because the test data indicates this format comes solely from the 949t properties file. If we discover this is not the case through qa testing, we can add the appropriate marc file fixtures and update the spec tests.

Fixes #303


awead:
@banukutlu Ok. I'll need to put an additional test in, if that's the case. Our tests were only covering the 949t case, and not the 006/008 case. That's easy to add though.

banukutlu:
> @banukutlu Ok. I'll need to put an additional test in, if that's the case. Our tests were only covering the 949t case, and not the 006/008 case. That's easy to add though.

Right, it looks like Instructional Material is also from 006/008 and yes would be nice to have an additional test for that case as well thought coverage still looks good but just for piece of mind maybe :)

banukutlu (changes requested):
overall refactor is great, seems like logic should be same but not so easy to review logic esp now that it has been a while I knew these on top of my head, tests are passing so all seems good. I guess other than Instructional Material spec, it looks great to me. Thanks @awead!

awead:
@banukutlu I've added additional tests for instructional materials

lib/psulib_traject/processors/record_type.rb
```diff
@@ -0,0 +1,76 @@
...
+        'Archives/Manuscripts'
+      else
+        Traject::TranslationMap
+          .new('formats_leader6')
+          .translate_array([record.leader[6]])
+          .first
+      end
+    end
+
+    def bibliographic_level
```
banukutlu, line 41: @awead Wondering if this could be private as well?

lib/psulib_traject/processors/record_type.rb
```diff
@@ -0,0 +1,76 @@
...
+        'Archives/Manuscripts'
+      when 'd'
+        book
+      when 'm'
+        thesis_or_book
+      end
+    end
+
+    # If decided that it is a Thesis/Dissertation, it is NOT a Book
+    def thesis_or_book
```
banukutlu, line 57: @awead same for this, could it be private?

lib/psulib_traject/processors/record_type.rb
awead: I thought about this too. Technically, everything other than `:resolve` should be private. Although the public/private distinction here gets a little weird because we're using a (the?) command pattern here, and the entry point for the whole thing is `RecordType.call`. Does that mean someone would:

``` ruby
type_processor = RecordType.new(record, current_formats)
type_processor.bibliographic_level
```

🤷  Probably not... And at the very least, it's not something we need to support right now. If we make them private, that will result in the tests doing `type_processor.send(:bibliographic_level)` which is considered "bad," but I'd rather have good test coverage.

Anyway, it's an interesting conundrum, and also the result of my chosen `.call` pattern here for the processors, which we could revisit...

awead:
@banukutlu I pushed up some new changes to make those method private. I alluded to this in my comments, but the only thing that needed to change was the test, i.e. using `:send`, which I think is completely acceptable in this case.

banukutlu (approved):
lgtm, thanks!

PR #307, opened 2021-06-28
docker compose

whereismyjetpack:
- adds docker-compose setup
- adds sidekiq to allow traject to be run as a sidekiq worker
- adds indexer worker classes as a thin wrapper around traject
- breaks up solr config and matches the connection string with psulib_blacklight
- allows the overlap file to be optional, logs a message when it can't be found

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ ARG UID=3000
 
 ENV BUNDLE_PATH=/app/vendor/bundle
 
-RUN apt-get update && apt-get install -y gcc
+RUN apt-get update && apt-get install --no-install-recommends -y gcc &&  \
+        rm -rf /var/lib/apt/lists*
 
 RUN useradd -u $UID app -d /app
 RUN mkdir /app/tmp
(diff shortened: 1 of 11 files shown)


whereismyjetpack:
This is very much a WIP, but wanted to get some feedback on some of the changes.

awead:
@whereismyjetpack can we try doing the docker-ization first without sidekiq as a first step? I feel like this biting off more that we can chew.

whereismyjetpack:
Sidekiq changes are fairly insignificant, and optional. I'd like to keep them so I can continue testing in k8s with sidekiq. We can keep this as a long running branch if we have to. The only breaking changes that I'm aware of are the shape of the config, so that it matches psulib_blacklight.

awead:
@whereismyjetpack I was suggesting that because your branches tend to get too long. It's much easier to help you if your PRs are smaller and more focused. Let me know when this is in a more complete state, and I can provide my comments at that point. If you want to make it easier to read, do force pushes. See multiple "revert" and "linting fixes" doesn't communicate anything significant to the reviewer. This is why I will often separate out large PRs into multiple commits so that the reviewer has an easier time comprehending.

PR #308, opened 2021-06-30
Adds sidekiq support

whereismyjetpack:
Adds sidekiq support, Docker is coming along with it as a second commit. some of the files bleed into one another. 
All current functions, aside from the solr configuration (I renamed collection_name to collection to match psulib_blacklight, and broke the URL up into parts to allow easier injection of variables from kubernetes. 
Adds traject rake tasks (these are additive to keep the current set working) that emulates what hourlies do, and what a full index would do as sidekiq workers. Indexing can be run with or without sidekiq, or can be run clasically via `bundle exec traject` 

TODO

- [ ] hourlies presently don't delete. I'm working on adding that as another PR. 
- [ ] hourlies aren't automated into the scratch space in kubernetes yet. i'm manually rsyncing these files to the traject pvc
- [x] CI/CD for shipping the container (i'm manually doing this for now)
- [ ] sidekiq UI for viewing progress. rake tasks for viewing progress, too perhaps.

I can address the items above as separate PRs or as part of this PR.


lib/psulib_traject/indexer.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'traject'
+require 'sidekiq'
+require 'sidekiq-scheduler'
+require 'redis'
```
awead, line 6: These should all move to `lib/psulib_traject.rb`

lib/psulib_traject/indexer.rb
```diff
@@ -0,0 +1,70 @@
...
+
+class HourliesWorker
+  include Sidekiq::Worker
```
awead, line 9: These should be under the PsulibTraject namespace, and each worker in its own file. You could define a base work class to include your worker. This like what Rails does.

``` ruby
module PsulibTraject
  class Worker
      include Sidekiq::Worker
  end
end
```

then,

``` ruby
module PsulibTraject
  class HourliesWorker < Work
  end
end
```
and so on....

Alternatively, there's `PsulibTraject::Workers::Base` and `PsulibTraject::Workers::Hourly`

Everything should be required in  lib/psulib_traject.rb

lib/tasks/traject.rake
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'traject'
+require './lib/psulib_traject/indexer'
+require 'sidekiq'
```
awead, line 5: this should just be `require 'psulib_traject'`

awead (changes requested):
Here are my initial thoughts. You can tack these on to the last commit as a force push, or just additional. If it helps we can pair up on this.

lib/tasks/traject.rake
whereismyjetpack: when changing this to `psulib_traject` it can't find it, when changing it to `./lib/psulib_traject` it finds it, but nothing within the modules require block loads. i.e `LoadError: no such file to load -- psulib_traject/hathi_overlap_reducer` I'm sure it's something i'm doing wrong with pathing or something, this is when running `bundle exec rake -T`

awead: I pushed up a commit to fix this. It goes in the Rakefile.

whereismyjetpack:
"settings and configuration change" commit is broken out because it's not related to the refactoring, but things I needed to do to make development easier. 

The second commit has in it parts of the refactor. I'm still working on this a bit, but wanted others to be able to see it

lib/psulib_traject/indexer.rb
whereismyjetpack: These have all been moved and namespaced in PsulibTraject module

awead (dismissed):

awead (approved):

PR #309, opened 2021-06-30
Use MarcBot

awead:
Installs the MarcBot gem and starts using it for some sample record creation.


banukutlu (approved):
awesome 🎉  looks great, thanks!

PR #324, opened 2021-08-04
updates lockfile for versions of jruby we're using in docker

whereismyjetpack:
I'm not sure what went sideways here, but syncing it up


awead:
My local jruby version is `jruby-9.2.11.1`. There is no `9.2.11.1-jdk14` version defined, so I think docker is using some different kind of nomenclature, perhaps determining both java and jruby version from the same string. Locally, I have java 11, which explains the discrepancy.

banukutlu (approved):
👍🏻

PR #325, opened 2021-08-04
adds rack config for sidekiq-web

### LINKED ISSUE #323: Add Sidekiq UI for viewing progress
Provide a UI for viewing Sidekiq workers.

## What does done look like

- Sidekiq Admin interface at `/sidekiq` from the catalog. This is webauth protected, only ul-dev-team can display it.

whereismyjetpack:
in k8s, we can run sidekiq-web at `/sidekiq` configure rack here. 

Fixes #323


lib/psulib_traject/workers/indexer.rb
```diff
@@ -6,10 +6,10 @@ class Indexer < Base
       # @param [Hash] args
       # @option args [String, Pathname] :filename (required)
       # @option args [String] :collection_name
-      def perform(filename, opts = {})
+      def perform(filename, collection_name = nil)
```
awead, line 9: was there a reason the hash parameter wouldn't work here? It's better to have optional parameters as a hash so you don't have to worry about positionality later on.

config.ru
```diff
@@ -0,0 +1,12 @@
+require 'sidekiq'
+require 'sidekiq-scheduler/web'
```
awead, line 2: sidekiq and sidekiq-scheduler are being required in psublib_traject. are these needed here?

config.ru
```diff
@@ -0,0 +1,12 @@
...
+
+Sidekiq.configure_client do |config|
+  config.redis = { :size => 1 }
+end
+
+require 'sidekiq/web'
```
awead, line 8: move this up with the other requires, or put them in publib_traject.rb.

awead (changes requested):
I have some questions about the require steps. We might need to move some around between this config.ru and psulib_traject.rb

lib/psulib_traject/workers/indexer.rb
whereismyjetpack: no idea, when called from rake the argument gets ignored. 

`bundle exec rake traject:index['foo','bar']`  bar gets ignored.

config.ru
whereismyjetpack: only the /web requires are in config now

whereismyjetpack:
@banukutlu I think this is ready now. I've re-arranged the requires, (imports?) and tested it on my preview branch, https://catalog-djb44.dev.k8s.libraries.psu.edu/sidekiq

banukutlu:
> @banukutlu I think this is ready now. I've re-arranged the requires, (imports?) and tested it on my preview branch, https://catalog-djb44.dev.k8s.libraries.psu.edu/sidekiq

kewl thanks, I'm on it 👍🏻

banukutlu (approved):
lgtm 👍🏻

lib/psulib_traject/workers/indexer.rb
awead: I'd still like to have an opts hash here. Which test was failing?

awead: OK, this is fixed now, but the rake tasks needed to be updated to reflect the different arguments.

awead (approved):

lib/psulib_traject/workers/indexer.rb
whereismyjetpack: I misspoke, _shocker_ i know. The issue was rake sends the argument as a string, and rspec sends the argument as a symbol, so when the tests were passing the method was ignoring the argument that the rake test was passing, i guess the rake task could do something like args.collection.to_sym? or something. feels hacky

awead:
@banukutlu My change to the rake tasks removed your previous approval, so I just went ahead, re-approved, and merged.

lib/psulib_traject/workers/indexer.rb
awead: Yeah, I just updated the rake tasks with the new positional arguments.

PR #326, opened 2021-08-04
Create forward and reverse LC shelf keys

### LINKED ISSUE #317: Generate and Index LC Shelf Keys
# To-Do
- [x] normalize call numbers using `lcsort` gem
- [x] use normalized call number as forward shelf key
- [x] generate reverse shelf key from forward shelf key, ex:
```return the reverse String value, mapping A --> 9, B --> 8, ... 9 --> A and also non-alphanum to sort properly (before or after alphanum)```
- [x] map the keys to `lc_forward_shelf_key` and `lc_reverse_shelf_key` fields

awead:
Using the Lcsort gem, this adds shelf keys that can be used to correctly sort LC call numbers in forward and reverse directions. This is used in conjunction with term component queries in the catalog to create lists of records ordered according to their call numbers.

Fixes #317


banukutlu (approved):
looks great! 🎉  thanks @awead

PR #332, opened 2021-08-09
Test Hourlies, CI

whereismyjetpack:
- Adds Tests for Hourlies Indexer. 
- Configures sidekiq-scheduler to run on the current schedule.
- Hourlies Job calls the right class now 
- When a file is indexed, it's added to redis semaphore, so it doesn't get indexed again. This allows another deployment to index the same files, or for us to easily replay a file without needing to sync the hourlies dir again.

Questions:
I wonder if we should make the cron schedule configurable?

TODO (as another PR)

- [ ] process deletes


spec/spec_helper.rb
```diff
@@ -10,6 +10,8 @@
 require 'rspec'
 require 'webmock/rspec'
 require 'psulib_traject'
+require 'rspec-sidekiq'
+require 'sidekiq/testing/inline'
```
awead, line 14: Put all of the changes to spec_helper into a file called `spec/support/sidekiq.rb`. It will be autoloaded from line 19. The spec_helper file is auto-generated when installing or updating rspec, so we minimize any additions to it to keep the diffs small.

spec/lib/psulib_traject/workers/hourly_indexer_spec.rb
```diff
@@ -1,14 +1,42 @@
...
   describe '#perform' do
     it 'submits jobs for each hourly file' do
-      # TBD
-      indexer.perform
+      indexer.perform_async
+      expect(described_class).to have_enqueued_sidekiq_job
+    end
+
+    it 'increases the size of the job queue' do
+      expect do
```
awead, line 27: We use { } on the do/end blocks here. I think it's because this is a lambda here? But I'm not sure that's 100% accurate.

spec/lib/psulib_traject/workers/hourly_indexer_spec.rb
```diff
@@ -1,14 +1,42 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
+require 'sidekiq/testing'
```
awead, line 4: I'd put this in the aforementioned `spec/support/sidekiq.rb` file.

spec/fixtures/hourlies/date_008.marc
```diff
@@ -0,0 +1 @@
...
 0aMass media and propaganda.
1 aChomsky, Noam.
423Contributor biographical informationuhttp://www.loc.gov/catdir/bios/random051/2001050014.html
423Publisher descriptionuhttp://www.loc.gov/catdir/description/random044/2001050014.html
  a2710183bHorizon bib#
  aBaker & TaylorbBKTYc18.95d14.21i0375714499n0003788716sactive
  aYBP Library ServicesbYANKn1813010
  aBaker and TaylorbBTCPn2001050014
  aP96.E25 H47 2002flcbelc1cc. 1q0i4659750lembluememsel
  aC0bJHE
```
awead, line 1: I don't see you loading these files anywhere. Am I missing something?

lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,13 +4,13 @@ module PsulibTraject
   module Workers
     class HourlyIndexer < Base
       def perform
-        target = Dir.glob("#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*.mrc")
+        target = Dir.glob("#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*.m*rc")
         indexed_files = redis.keys('hr:*').map! { |e| e.gsub('hr:', '') }
```
awead, line 8: You're calling `map!` on the array, which means you'd be modifying the actual array, and overwriting the elements within `redis.keys`. I'm not sure that's what you want... maybe just `map` here since indexed_files is capturing the new output?

spec/lib/psulib_traject/workers/hourly_indexer_spec.rb
```diff
@@ -1,14 +1,42 @@
...
 
-RSpec.describe PsulibTraject::Workers::HourlyIndexer, skip: ci_build? do
-  let(:indexer) { described_class.new }
+Sidekiq::Testing.fake!
```
awead, line 6: I'm wondering if this shouldn't live in `spec/support/sidekiq.rb` too. Does it need to be "unfaked" at some point?

awead (changes requested):
Looks good! rspec-sidekiq is a nice addition. We should add that to our other repos too.

spec/lib/psulib_traject/workers/hourly_indexer_spec.rb
whereismyjetpack: This, I'm not sure. it looks super weird to me, but was yanked from the rspec examles. the `{}` syntax works fine, I think I got it right.

spec/fixtures/hourlies/date_008.marc
whereismyjetpack: These get loaded here. (https://github.com/psu-libraries/psulib_traject/blob/a8cd1765848ffec023ac5033e7a3450437bfe48f/spec/lib/psulib_traject/workers/hourly_indexer_spec.rb)

The tests ensure that both marc, and mrc files are read, and that exactly 2 jobs are put on the queue.

lib/psulib_traject/workers/hourly_indexer.rb
whereismyjetpack: Good catch. Thanks, StackOverflow.

spec/spec_helper.rb
whereismyjetpack: added.

spec/lib/psulib_traject/workers/hourly_indexer_spec.rb
whereismyjetpack: I'd think if it needs to be unfaked we can re-arrange at that point.

awead (approved):

banukutlu:
> * Adds Tests for Hourlies Indexer.
> * Configures sidekiq-scheduler to run on the current schedule.
> * Hourlies Job calls the right class now
> * When a file is indexed, it's added to redis semaphore, so it doesn't get indexed again. This allows another deployment to index the same files, or for us to easily replay a file without needing to sync the hourlies dir again.
> 
> Questions:
> I wonder if we should make the cron schedule configurable?
> 
> TODO (as another PR)
> 
> * [ ]  process deletes

configurable cron would be nice actually, not normally but rarely we changed the cron for example during ETAS switch, a big extract from Symphony took longer, so we ended up changing the cron. Such thing only happens very rarely so it would be a nice thing to have but not required.

banukutlu (approved):
lgtm 👌🏻 thanks @whereismyjetpack!

Issue #285, opened 2021-04-15
Identify Base Call Numbers


ruthtillman:
https://github.com/psu-libraries/psulib_blacklight/wiki/Browse-Items-by-Library-of-Congress-Call-Number#identify-base-call-numbers

# To-Do
- [x] #316  
- [x] #318

banukutlu:
@ruthtillman  Here are some preliminary call number base results for sample record (not the examples you provided yet.)

[call_number_base_sample.txt](https://github.com/psu-libraries/psulib_traject/files/6718857/call_number_base_sample.txt)

Note that this is not finalized, only tries to find base numbers for LC and Dewey not the other classifications yet.

ruthtillman:
Review notes (running comment, will edit with additional updates):

1. While this pertains to ASIS, worth revisiting when we do ASIS and checking Regex to be sure it's supported -- on catkey: 55858 we have weird dates like `Jan.1-15 1973`. There's no space between the month abbreviation, the period, and the date range. Also the dates are a range. CAT handles this by breaking it down item-by-item  which isn't really helpful.

PR #333, opened 2021-08-10
Return explicit shelf key types in struct

awead:
The keymap_struct needs to indicate which key is the forward and reverse key. This will allow to us build the correct list when retrieving them from Solr in the catalog.

Related to #285


banukutlu (approved):
👌🏻

PR #334, opened 2021-08-10
NullKey for unparseable LC call numbers

awead:
If we can't determine a proper shelf key, we need to return nil. Otherwise, querying on shelf keys that have bad characters in them will cause errors in Solr

I've tested this with our catalog's call number branch, and this doesn't seem to break anything, and it should fix the errors we're seeing with Solr queries. All of the LC numbers that have bad characters in them, ex (?:, are all in call numbers that Lcsort can't process. Returning nothing is the better option here. Later, we can attempt to try our own process for this call numbers.

Related to #285


banukutlu (approved):
lgtm 👌🏻

PR #337, opened 2021-08-12
Adds deletes to hourlies

whereismyjetpack:
- runs hourlies in serial (in testing this is much faster, due to the startup cost of jruby)
- runs deletes after hourlies (another reason to run indexing serially, deletes must happen after indexes)


lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,15 +4,40 @@ module PsulibTraject
   module Workers
     class HourlyIndexer < Base
       def perform
-        target = Dir.glob("#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*.m*rc")
+        target = Dir.glob(
+          "#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*.m*rc"
```
awead, line 8: I would do something like: 

``` ruby
def hourlies_directory
  @hourlies_directory ||= Pathname
                            .new(ConfigSettings.symphony_data_path)
                            .join(ConfigSettings.symphony_hourlies_subdir)
end
```

Then here do
``` ruby
target_files = hourlies_directory
  .children
  .select { |filename| filename.extname.match?(/ma?rc/' }

File: lib/psulib_traject/workers/hourly_indexer.rb [hunk 2]
```diff
@@ -4,15 +4,40 @@ module PsulibTraject
...
         files_to_index.each do |t|
-          puts "indexing #{t}"
-          PsulibTraject::Workers::Indexer.perform_async(t)
+          indexer.logger.info "marking #{t} as done"
           redis.set("hr:#{t}", true)
         end
+
+        deletes = Dir.glob(
+          "#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*_deletes_*.txt"
+        )
```
awead, line 25: Use the memoized directory here, and match on deletes.

lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,15 +4,40 @@ module PsulibTraject
...
         indexed_files = redis.keys('hr:*').map { |e| e.gsub('hr:', '') }
         files_to_index = target - indexed_files
 
+        indexer.logger.info "Found #{files_to_index.length} files to index"
+        array_of_files = files_to_index.map { |file| File.new(file) }
+
+        indexer.process array_of_files
+
         files_to_index.each do |t|
```
awead, line 18: use a complete word here. "t" isn't clear.

lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,15 +4,40 @@ module PsulibTraject
...
+        files_to_process = deletes - processed_deletes
+
+        indexer.logger.info "Found #{files_to_process.length} files to process for deletes"
+
+        files_to_process.each do |file_name|
+          File.open(file_name, 'r') do |f|
+            f.each_line do |line|
+              id = line.chomp
+              indexer.writer.delete(id)
+              indexer.logger.info "Deleted ID #{id}"
```
awead, line 36: You can simplify this to:

``` ruby
File
  .read(file_name)
  .split("\n")
  .map { |id| delete(id) }

def delete(id)
  indexer.writer.delete(id)
  indexer.logger.info "Deleted ID #{id}"
end
```

lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,15 +4,40 @@ module PsulibTraject
...
       def perform
-        target = Dir.glob("#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*.m*rc")
+        target = Dir.glob(
+          "#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*.m*rc"
+        )
         indexed_files = redis.keys('hr:*').map { |e| e.gsub('hr:', '') }
         files_to_index = target - indexed_files
 
+        indexer.logger.info "Found #{files_to_index.length} files to index"
+        array_of_files = files_to_index.map { |file| File.new(file) }
```
awead, line 14: Pathname.children will return an array of Pathname objects, which wrap calls to File, so you won't need to re-init each one as its own File.

lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,15 +4,40 @@ module PsulibTraject
...
+          indexer.logger.info "marking #{t} as done"
           redis.set("hr:#{t}", true)
         end
+
+        deletes = Dir.glob(
+          "#{ConfigSettings.symphony_data_path}/#{ConfigSettings.symphony_hourlies_subdir}/**/*_deletes_*.txt"
+        )
+        processed_deletes = redis.keys('hr:*').map { |e| e.gsub('hr:', '') }
+        files_to_process = deletes - processed_deletes
+
```
awead, line 28: This method is getting long. I would put all the deleting in a separate method, or even a different worker so we can call the deleting independently of the indexing, if we need to.

awead (changes requested):
Needs some clean up, but it's all there. I'm suggesting using a new `Workers::Deleter` class, or at least a separate method on HourlyIndexer.

lib/psulib_traject/workers/hourly_indexer.rb
whereismyjetpack: oh, i didn't realize that .children would do the same thing as `/**/*` thanks.

lib/psulib_traject/workers/base.rb
```diff
@@ -9,6 +9,12 @@ def self.perform_now(*args)
         new.perform(*args)
       end
 
+      def indexer
+        @indexer ||= Traject::Indexer::MarcIndexer.new
+        @config ||= @indexer.load_config_file('config/traject.rb')
+        @indexer
```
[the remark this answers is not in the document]
whereismyjetpack: oh, that's neat. i didn't realize you could do that.

whereismyjetpack:
I think i've addressed all the concerns!

lib/psulib_traject/workers/hourly_indexer.rb
```diff
@@ -4,14 +4,56 @@ module PsulibTraject
...
+          .select { |filename| filename.extname.match?(/ma?rc/) }
+
         indexed_files = redis.keys('hr:*').map { |e| e.gsub('hr:', '') }
-        files_to_index = target - indexed_files
+        files_to_index = target_files.reject { |filename| indexed_files.include?(filename.to_s) }
+
+        indexer.logger.info "Found #{files_to_index.length} files to index"
+        array_of_files = files_to_index.map { |file| File.new(file) }
+
+        indexer.process array_of_files
```
awead, line 52: Did you try using `files_to_index` here? It should work because Pathname wraps all the functions of File.

awead (changes requested):
One last question, otherwise it's good to do.

lib/psulib_traject/workers/hourly_indexer.rb
whereismyjetpack: I missed this, I just tried it, and marc4j_reader throws the following 

       undefined method `to_inputstream' for <Pathname:./spec/fixtures/hourlies/date_008.marc>:Pathname

awead: Ah ok, I think this might be a weird thing with jruby.

awead (approved):

PR #340, opened 2021-08-17
Custom shelf keys for LC call numbers

awead:
When Lcsort is not able to create a shelf key, we manipulate the call number into something that it can sort. Lower case characters are converted to numbers, colons are removed, and a zero is added to the LC classification if no number is present.

### Lower-Case Cutters

If a call number has lower-case values in the cutter, ex. `PZ7.H56774Fou 2014` the key will look like `PZ.0007.H56774.F1420--2014` where the "ou" has been translated into "1420". This way, the key should still sort. Cases like `LD4481.P8kG45` produce a key that looks like `LD.4481.P810.G45--2008`, where the "k" is replaced with "10".

### LC Classifications with no digits

If the LC classification part of the call number has no digits, ex. `KKP.B634b`, we add a zero. This assumes that "KKP" would sort _before_ "KKP1" or "KKQ". So the preceding example has a key that looks like `KKP0000.B63401`, with padded zeros and the "b" in the cutter translated to "01".

### DIFF
--- a/lib/psulib_traject/shelf_key.rb
+++ b/lib/psulib_traject/shelf_key.rb
@@ -5,6 +5,9 @@ module PsulibTraject
     FORWARD_CHARS = ('0'..'9').to_a + ('A'..'Z').to_a
     CHAR_MAP = FORWARD_CHARS.zip(FORWARD_CHARS.reverse).to_h
 
+    # Map lower-case letters to numbers for cutter sorting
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
looks good to me 👌🏻

ruthtillman:
The first set of logic looks correct, I would think the numbers would handle it well. And assumption: `This assumes that "KKP" would sort before "KKP1" or "KKQ". ` is correct!

ruthtillman (approved):

Issue #316, opened 2021-07-23
Identify Base LC Call Numbers


banukutlu:
Part of #285 

# To-Do
- [x] Lop LC call numbers to identify the base
- [x] Map the base call number to `call_number_lc_ssm` field

# Questions
 - Will the un-lopped call number displayed as well? See https://github.com/psu-libraries/psulib_traject/issues/316#issuecomment-885715444

ruthtillman:
No, we won't want to display unlopped. Those only matter in the actual record which we get from WebServices availability. We'll just want to display the bases in a sorted order.

awead:
Original comments by @ruthtillman 

In general: This is looking really good and very exciting to see! 

Regex issues and new Regex to add:
- [x]  We need to add support for `k.1`, `k.2` and `h.1, h.2` etc to regex for vol_parts (so basically standalone "k" and "h"). Example records: 3752257, 3752276. These are part-types I hadn't encountered before! Testing sample ++
- [x]  Catkey 1443986: Really weird call number formatting with stuff like `50th YYYY`.  _Expected:_ `TK7872.O7S901`
  - [x]  Related: 79172. Need to add support for these `\d{1,3}th` (22nd, 33rd) cases. _Expected_: `HV1796.P3`
- [x]  Catkey 1793712: mostly good except `HQ1101.M72 Index Spr.1972-Feb.1974` shows up alongside correct base call numbers. Unsure what's causing that, but it's a really weird call no. _Expected_: `HQ1101.M7214 | HQ1101.M72`  (there are 2 base call nos expected)
- [x]  Catkey 175885: it's not stripping the year for some reason. _Expected_: `HD1761.F37`
- [x]  Catkey 12133277: can't figure out what's causing third base call no: `PN1997.A1M216 2014 D`. (Conversely, 22083281 and 22083276 are working fine! what's the difference?) _Expected_: ` PN1997.A1M216 2014 DVD | PN1997.A1M216 2014 Blu-ray`
  - [x] Related: 21588551. _Expected_: `PN1997.2.L66255 2017 DVD | PN1997.2.L66255 2017 Blu-ray/DVD | PN1997.2.L66255 2017 Blu-ray`
  - [ ] Related: 2208330. _Expected_: `PN1997.2.V5454 2017 Blu-ray/DVD`
[...]

PR #342, opened 2021-08-20
Index LC Call Numbers

awead:
Extracts "base" call numbers from LC numbers in each record, and creates both forward and reverse shelf keys that are used in Solr to build a call number browse feature in the catalog.

Ref #316 
Ref #317


awead:
@banukutlu Note, I don't think we should squash this, since the commits are pretty specific and point to different tickets.

banukutlu (approved):
lgtm 👍🏻

PR #346, opened 2021-08-27
Ignore any call numbers that contain "Newspaper"

### LINKED ISSUE #341: Anything with "Newspaper" isn't good
Ignore any LC call number with "Newspaper" in it.

## WDLL

Change `value == 'Newspaper'` to `value.match?(/Newspaper/)`

nicktiberi:
Fixes #341.

### DIFF
--- a/lib/psulib_traject/call_number.rb
+++ b/lib/psulib_traject/call_number.rb
@@ -45,7 +45,7 @@ module PsulibTraject
     end
 
     def newspaper?
-      value == 'Newspaper'
+      value.match?(/Newspaper/i)
     end
 
     def local?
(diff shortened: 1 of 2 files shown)


ruthtillman:
Question: what does ` { is_expected.not_to be_periodical }` mean, functionally, for testing? 

I ask because technically many newspapers are, literally "periodicals" and might meet indexing criteria for Journal/Periodical... however if it's a behavior thing related to how we're filtering out the word "Periodical" that's fine!

awead:
@ruthtillman This is a Rails/RSpec-ism. In ruby, you can define methods that return booleans such as:

``` ruby
class Something
  def thing?
     true
  end
end
```

This is purely convention. If it ends in a question mark, it _should_ return true/false. RSpec piggybacks on this and lets you write a test like:

``` ruby
expect(instance_of_something).to be_thing
```

RSpec interprets `be_thing` to be a method called `thing?`

I think we call this "syntactic sugar" so it makes Ruby look cooler, but it's really just enabling lazy programmers.

ruthtillman:
Great, thanks!

Issue #329, opened 2021-08-09
Generate and index shelf keys for browsing subjects


banukutlu:
Browsing subjects can just be a facet. Doesn't need anything fancy, but, it needs its own field that's different from the current subject facets.

## To-Do

Create a `subject_browse_facet` field for subject browse:

     650|*0|abcdgvxyz:650|*1|abcdgvxyz:650|*3|abcdgvxyz:650|*3|abcdgvxyz`
     
     If `'650|*7|2'.downcase == 'pst'`, use `650|*7|abcdgvxyz`

For more details https://github.com/psu-libraries/psulib_traject/issues/329#issuecomment-906591131

~~Normalize subjects and generate forward shelf keys~~
~~Generate reverse shelf keys from forward shelf keys~~
~~Map the keys to `subject_forward_shelf_key` and `subject_reverse_shelf_key` fields~~

awead:
## Alternative Implementation

There's a way we may be able to implement this using the existing facets. We can tweak Blacklight's existing facet navigation and get something like this:

https://catalog-call-number-browse.dev.k8s.libraries.psu.edu/catalog/facet/subject_facet?facet.page=10&facet.prefix=American&facet.sort=index

![Screen Shot 2021-08-11 at 10 15 32 AM](https://user-images.githubusercontent.com/312085/129045491-ab3066b3-07f0-4176-9b92-2fa0eb1415b4.png)

The `facet.prefix` parameter is your "left-anchored search" parameter. We would remove the _Numerical Sort_ option.

### Advantages

1. Super easy.
2. Less complicated.
3. Ready to implement now
4. No changes required in traject

### Drawbacks

1. Could be slow. The forward/reverse key solution might be more performant
2. Doesn't have all the features?
3. More complicated, more work
4. Left-anchored search issues

For example, if I search for "American fiction", I'm going to get a listing of only the subject headings that _start_ with "American fiction" and no others. So, I'm browsing through all of these, when I get to the end it looks like:

https://catalog-call-number-browse.dev.k8s.libraries.psu.edu/catalog/facet/subject_facet?facet.page=17&facet.prefix=American+fiction&facet.sort=index

I won't see what comes after that alphabetically, like "American friction" (which doesn't exist as a subject heading, I'm just making that up as an example).

PR #347, opened 2021-08-27
Add subject_browse_facet

awead:
Addresses #329 

Creates a new `subject_browse_facet` field to support browsing subjects. Includes a refactoring of our existing macros, which splits out the subject-related macros into their own file, then refactors the tests and macro code separately.


banukutlu (approved):
👌🏻

PR #349, opened 2021-09-01
fix for subjects when record with 650 has an indicator but no subfields

banukutlu:
mapping was failing for records with weird 650s

eg. https://catalog.libraries.psu.edu/catalog/14651589/marc_view

Screen Shot 2021-09-01 at 1 53 56 PM

### DIFF
--- a/lib/psulib_traject/macros/subjects.rb
+++ b/lib/psulib_traject/macros/subjects.rb
@@ -14,7 +14,10 @@ module PsulibTraject::Macros::Subjects
 
       subjects = []
       Traject::MarcExtractor.cached(fields).collect_matching_lines(record) do |field, spec, extractor|
-        subjects << extract_subjects(field, spec, extractor).join(SEPARATOR)
+        extracted_subjects = extract_subjects(field, spec, extractor)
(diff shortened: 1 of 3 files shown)


nicktiberi (approved):
Seems like a logical fix to me! 👍

awead:
Alternatively, we could have our own implementation of `join`:

``` ruby
> SEPARATOR = " "
> [].inject { |memo, word| memo + SEPARATOR + word } 
=> nil
> ['adsf'].inject { |memo, word| memo + SEPARATOR + word } 
=> "adsf"
> ['adsf', 'abr'].inject { |memo, word| memo + SEPARATOR + word } 
=> "adsf abr"
```

banukutlu:
> yeah. the fact that ruby does:
> 
> ```
> > [].join('wtf')
> => ""
> ```
> 
> Doesn't quite sit right with me, but I guess because `.join` has to return a string and the best way to map nil to string is "". This also highlights a problem with traject that _everthing_ is an array, and it gets confusing.

yesss!!! srsly annoying

> > [].join('wtf')
> => ""

PR #355, opened 2021-09-07
Select only the full subject headings for browse

### LINKED ISSUE #354: Change subject browse facet indexing/display behavior to only use fullest form
For regular subject_facet, we have methods to break it out into all its various forms. Therefore on a record with

`Quilting—Pennsylvania—Cumberland County—History—19th century`

we also process it so we can generate links for:

```
Quilting—Pennsylvania—Cumberland County—History
Quilting—Pennsylvania—Cumberland County
Quilting—Pennsylvania
Quilting
```

However, for subject browse facet, we'll want to only use the fullest form of each subject we index, e.g. `Quilting—Pennsylvania—Cumberland County—History—19th century` Otherwise a person would see this;

Quilting—Pennsylvania—Cumberland County 	1
Quilting—Pennsylvania—Cumberland County—History 	1
Quilting—Pennsylvania—Cumberland County—History—18th century 	1

reasonably think that they have a bunch of different resources here and has no way of knowing they're all the one book.

So for each 650, subject browse facet should only have the longest form.

awead:
Uses a SubjectHeading class that can compare the size of a subject heading by how many sub-headings it has. This is used to create browseable subject headings that only contain the complete headings and not any intermediary ones.

Related #354


banukutlu (approved):
👌🏻

PR #356, opened 2021-09-09
Adds integration test for browseable subjects

### LINKED ISSUE #354: Change subject browse facet indexing/display behavior to only use fullest form

awead:
This also includes a refactoring to extract out helper methods that we can share across multiple tests.

Fixes #354


awead:
Whoops... we don't need this...

PR #360, opened 2021-09-27
Use shelvit gem for generation of normalized shelf keys

nicktiberi:


lib/psulib_traject/shelf_key.rb
```diff
@@ -20,7 +17,7 @@ def initialize(call_number)
 
     # @return [String]
     def forward
-      Lcsort.normalize(call_number) || normalize
+      Shelvit.normalize(call_number) || NullKey.new
```
awead, line 20: I wonder if NullKey should be incorporated into Shelvit. I don't think it has to be done now, but could be ticketed for future discussion.

awead (approved):

banukutlu (approved):
lgtm 👌🏻

PR #369, opened 2021-11-29
add more fields to subject browse

### LINKED ISSUE #359: Add a few fields (still smaller than normal subject) to subject_browse
Following up from testing, we should re-add these fields to the subject_browse_facet. These are fairly limited forms (in terms of which letters we're using, this means we can find all books _about_ Karl Marx and his work, for example) and restricted to LC so they should be pretty good for grouping the types of things without leading to too much spread, which we were getting with the original pass.

people:

`600|*0|abcd`
`610|*0|abcd`
`611|*0|abcd`

geography:

`651|*0|agvxyz`

named events and uniform titles:

`630|*0|avxyz`
`647|*0|avxyz`

Or basically just put in:

`600|*0|abcd:610|*0|abcd:611|*0|abcd:651|*0|agvxyz:630|*0|avxyz:647|*0|avxyz`

banukutlu:
Closes #359


config/traject.rb
```diff
@@ -255,7 +255,7 @@
 
 ## Subject browse facet
 to_field 'subject_browse_facet', process_subject_browse_facet(
-  standard_fields: '650|*0|abcdgvxyz:650|*1|abcdgvxyz:650|*3|abcdgvxyz:650|*3|abcdgvxyz',
+  standard_fields: '600|*0|abcd:610|*0|abcd:611|*0|abcd:630|*0|avxyz:647|*0|avxyz:650|*0|abcdgvxyz:650|*1|abcdgvxyz:650|*2|abcdgvxyz:650|*3|abcdgvxyz:651|*0|agvxyz',
```
awead, line 258: Do we need a test to make sure we don't get any regressions? Doesn't have to be much... use the existing ones, or write one specifically for this change.

awead (changes requested):

config/traject.rb
banukutlu: I thought about it as well but was not sure looking at the existing ones but I can def. add sth 👍🏻

banukutlu: specs added

lib/psulib_traject/macros/subjects.rb
```diff
@@ -35,24 +35,28 @@ def process_subject_browse_facet(standard_fields:, pst_fields:)
...
 
       accumulator.replace(
         subjects
-          .select { |subject| subject.length == subjects.max.length }
-          .map(&:value)
-          .uniq
+          .group_by(&:tag)
+          .map do |_tag, subjectheadings|
+            subjectheadings
+              .select { |subject| subject.length == subjectheadings.max.length }
```
awead, line 56: I'm not sure I understand how this is working. `subject.length` is now a string. Previously, it was the count of the number of items in the array of headings. We should be returning the subject heading that has the most number of elements, not the one that's the longest.

awead (changes requested):
I'm not sure this is doing what we need it to do. It might help if we used actual subject heading terms. Do we have a sample record in the catalog to test this against?

lib/psulib_traject/macros/subjects.rb
banukutlu: hmm OK, I will fix that, I misunderstood that part 👍🏻

banukutlu: I changed the length method to how it was before, I think that kinda did it 🤔

awead (approved):

PR #370, opened 2021-12-01
Index summary holdings as json struct

### LINKED ISSUE #348: Summary holdings indexing
# Indexing Summary Holdings

First, what are summary holdings? Summary holdings provide information "at a glance" regarding our journals, magazines, newspapers, and other periodicals. For example, if we have over 100 volumes, the summary holdings could provide you with the basic, helpful information: 1893-1922; 1931-1935; 1937-1992; Indexes to 1980-1990. --- which helps you know you won't find anything more recent than 1992 and there are some gaps.

It might also explain what a call number like "Periodical" means -- e.g. that we have "Current issue" or "Current year only."

## What Summary Holdings Consist Of

Summary holdings consist of "units" with some combination of the following 3 fields:

* 852 - the location data and call number of these holdings (required)
* 866 - the actual holdings information (common)
* 867 - holdings information about supplemental material (uncommon)
* 868 - holdings information about indexes (uncommon)

| Field | subfield | label | expectation | notes |
| --- | --- | --- | --- | --- |
| **852** | a | n/a | United States library code | we can ignore this because it's at the level of "Penn State |
|  | b | Library | _our_ library code | e.g. `YORK` or `NEWKEN` or `UP-ANNEX` |
|  | c | Location | _our_ location code | e.g. `CATO-2` or `STACKS-YK` or `MFILM-NML` |
|  | hi | Call Number | call number (essentially text string) | call number associated with the holdings, such as `Microfilm E158` or `Periodical` - the 852 may or may not contain an "i" but the "i" should just be indexed for display in conjunction with the "h" -- be sure that spacing looks right. |
| **866** | a | Holdings Summary | display information for holdings | a long text string | 
| **867** | a | Supplemental | display information for supplement holdings | a long text string |
| **868** | a | Index | display information for holdings of indexes | a long text string |

(text strings may be _quite_ long as repetition is based on a difference in the material, not the fact that a list is going on _forever_ -- so you'll see repetitions here which are "supplemental filmings," for example, which are probably duplicates.)

Unfortunately, [summary holdings are weird](https://github.com/psu-libraries/psulib_blacklight/wiki/Holdings-Indexing#why-summary-holdings-are-weird).

## What We Can Expect

These are things we should be able to expect when parsing holdings:

1. Each holding "unit" will begin with an 852 which ties it to a specific location. (1,1)
2. Each holding "unit" may contain one or more 866s (0,~)
3. Each holding "unit" may contain one or more 867s (0,~)
4. Each holding "unit" may contain one or more 868s (0,~)
5. Within each holding "unit" the MARC fields will be in order (so we could test for each one's existence if there's more than one 852)

It is _unlikely_ that a unit will be just 852 and 867 or that it will contain repeating 867s or 868s, but we should not rule out these possibilities. Repetition is primarily based around any differentiation in the data, not length. Note: I don't have proof that 867 or 868 ever repeat within a unit, I just also don't have proof that they don't--we could analyze the data if that would be helpful, but since 866 can also repeat it makes about as much sense just to plan for three possibly-repeating fields.

## Sample Catkeys:

all contain required 852s. 867 and 868 examples will also contain 866 (and may also contain 868/867 respectively)

* Contain 866: 1846251, 2702238, 762960, 811044, 483004
* Contain 867: 1681275, 1979901, 1994409, 2004296, 2065914, 15417033, 15417373, 2701356, 2701384, 3276182, 2701588, 2701834
* Contain 868: 1199576, 821732, 189479

See wiki for more examples and documentation: https://github.com/psu-libraries/psulib_blacklight/wiki/Holdings-Indexing

awead:
Collects summary holdings data and stores it in an array of json hashes. These will be parsed by the catalog and presented in the record view.

Fixes #348


spec/lib/psulib_traject/processors/summary_holdings_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        expect(holdings.index).to contain_exactly('Index')
+      end
+    end
+
+    context 'when multiple summary holdings are present' do
+      let(:record) { MARC::Reader.new(File.join('./spec/fixtures/summary_holdings.mrc')).to_a.first }
+
+      its(:count) { is_expected.to eq(3) }
+    end
+  end
```
banukutlu, line 38: @awead Could we also test for when there are multiple 852s? Another potential one maybe to test when there is a 852 with no summary followed by a 852 with summary?

banukutlu (changes requested):
This looks great, just one comment re for couple more tests maybe?

spec/lib/psulib_traject/processors/summary_holdings_spec.rb
awead: Sure thing!

spec/lib/psulib_traject/processors/summary_holdings_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context 'when multiple summary holdings are present' do
+      let(:record) { MarcBot.build(:multiple_summary_holdings) }
+
+      it 'returns two Unit objects' do
+        first_holdings = summary_holdings[0]
+        expect(first_holdings).to be_a(PsulibTraject::PeriodicalHoldings)
+        expect(first_holdings.library).to eq('UP-ANNEX')
+        expect(first_holdings.location).to eq('CATO-2')
+        expect(first_holdings.call_number).to eq('HX1 .M3')
```
banukutlu, line 35: Can we also check if summary, supplement and index is empty here? 🤔 

```suggestion
        expect(first_holdings.call_number).to eq('HX1 .M3')
        expect(first_holdings.summary).to be_empty
        expect(first_holdings.supplement).to be_empty
        expect(first_holdings.index).to be_empty
```
```

banukutlu (changes requested):
@awead  Thanks for adding the additional specs, I have one more minor comment.

spec/lib/psulib_traject/processors/summary_holdings_spec.rb
awead: yep yep. all done. but, @nicktiberi needs some changes to the data structure, so there will be further changes.

banukutlu (dismissed):
👌🏻

spec/lib/psulib_traject/processors/summary_holdings_spec.rb
banukutlu: ok, I approved but we can hold off on merging 👍🏻  thanks!

banukutlu (approved):
LGTM! 👍🏻

nicktiberi:
We may need to slightly modify this to index all the subfields of 866, 867, 868 per #373. This came up on Friday when Ruth and I were looking at a record in the CAT.

PR #376, opened 2021-12-08
Use any a-z subfield for summary holdings information

### LINKED ISSUE #373: Change 866, 867, 868 subfields to ATOZ (traject method)
Testing with Nick uncovered the possibility of other fields being used in the data, particularly `|z`. Since these fields are not thoroughly-documented, it seems like a good use case for the traject method "ATOZ" instead of just listing numbers.

This will index any subfields (that use letters) and, as usual, in the order in which they appear (which is the order the person entering data intended since that's how traditional MARC displays).

Each field should still go to a single string as this is for display only. We may need to test to be sure that spaces between them are appearing correctly.

awead:
866, 867, and 868 fields have summary holdings information in subfields other than just |a, so we need to index all these subfields.

Fixes #373

### DIFF
--- a/lib/psulib_traject/periodical_holdings.rb
+++ b/lib/psulib_traject/periodical_holdings.rb
@@ -42,7 +42,9 @@ module PsulibTraject
     end
 
     def summary
-      @summaries.map { |field| field['a'] }
+      @summaries
+        .map { |field| atoz(field) }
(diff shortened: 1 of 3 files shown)


nicktiberi (approved):
LGTM! 🚀

PR #378, opened 2021-12-14
Index 995 fields for dedication info keyword search and display

### LINKED ISSUE #12: Mapping for digital bookplates and dedications
Ruth updated 2021-11-16 with current practices and mapping so things below may apply less. We will currently not display actual digital bookplates, but the data we're creating through indexing will let us create image bookplates/enhanced search/however we want to do it. See also wiki: https://github.com/psu-libraries/psulib_blacklight/wiki/Endowments-and-Dedications and https://github.com/psu-libraries/psulib_blacklight/issues/57 for display.

Development and selectors need to be able to search for and collocate materials purchased on specific funds. We currently display these here: https://libraries.psu.edu/about/give-libraries/endowments/700000010161 and they link to the CAT. While they link to the 541 for search, [long sidebar about how indexes work in the CAT] tl;dr it's actually the 993 which is being folded into a search index _called_ 541. We can improve the matching by searching the 993a, specifically.

We should display the name of the fund in the record with a link to the fund code search. The fund's name should be in the keyword search so that more casual users could find it.

We also use the 995 for dedication values and while I'm not entirely clear about their source, we need to index them for display and they should be keyword searchable.

## MARC Fields

We do not use indicators, just subfields.

| MARC subfield | description | expected value | use for |
| --- | --- | --- | --- |
| 993a | fund code for the acquisition source | a string | controlled search |
| 993n | fund name | text | keyword search, display |
| 995acn3 | dedication value | text | keyword search, display |

## Sample Catkeys

993: `4322824, 33838462, 28901656, 34853690, 8049243, 33179445`

995: `33839158, 2872383, 1510037, 3343812`

### 4322824: repeated field

(also `33179445`)

```
993 // a| 700000010161 n| Allison-Shelley Collection Fund
993 // a| 700000010275 n| Paterno Libraries Endowment (Campus College Libraries)
```

### 33839158: Dedication - basic

```
995 // a| In Memory of Dr. David S. Bender, Associate Professor of Educational Psychology, 43+ Years of Service.
```

### 2872383: Dedication - complex

```
995 // c| Selected to honor a| Mr. Steve Borelli, on the occasion of receiving tenure-promotion; and purchased with funds from the a| Paterno Libraries Endowment; d| 2020. 
``

### PR DESCRIPTION
**Pull request by nicktiberi**
Fixes #12.

### DIFF
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -587,6 +587,9 @@ to_field 'location_facet', extract_marc('949l'), exclude_locations, translation_
 # 993 - Endowment Information
 to_field 'endowment_info_note_ssim', extract_marc('993an')
 
+# 995 - Dedication Information
+to_field 'dedication_info_note_ssim', extract_marc('995acn3')
+
 # Serials fields
 #
 # Preceding and Succeeding Entries display
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -588,7 +588,7 @@ to_field 'location_facet', extract_marc('949l'), exclude_locations, translation_
 to_field 'endowment_info_note_ssim', extract_marc('993an')
 
 # 995 - Dedication Information
-to_field 'dedication_info_note_ssim', extract_marc('995acn3')
+to_field 'dedication_ssim', extract_marc('995acn3')
 
 # Serials fields
 #

### CODE REVIEW

File: config/traject.rb [hunk 1]
```diff
@@ -587,6 +587,9 @@
 # 993 - Endowment Information
 to_field 'endowment_info_note_ssim', extract_marc('993an')
 
+# 995 - Dedication Information
+to_field 'dedication_info_note_ssim', extract_marc('995acn3')
```
banukutlu, line 591:
```suggestion
to_field 'dedication_ssim', extract_marc('995acn3')
```

what do you think for a shorter field name?

banukutlu (changes requested):

config/traject.rb
nicktiberi: I'm totally fine with that. I guess I was just going off of the convention for other fields with the `_note` suffix, but I feel like that is kind of unnecessary. `dedication_ssim` is cool with me. 🆒

banukutlu: yeah we just need the `_ssim` suffix, endowment field could have just been `endowment_info_ssim` but does not matter here. thanks for changing :)

banukutlu (approved):
👍🏻

PR #379, opened 2021-12-14
changing rules for full and partial links so to prevent colliding urls

### LINKED ISSUE #301: Review full and partial links rules
Full links available? rule:
`(ind2 == '0' || ind1 == '4' || (ind1.strip.empty? && ind2.strip.empty?)) && !NOT_FULLTEXT.match?(url_label)`

Partial links available? rule:
`ind2 == '1' || ((ind2 == '0' || ind1 == '4') && NOT_FULLTEXT.match?(url_label))`

856s like the below are both full and partial:
`856 4 1 3| v.19-20 Sept.1939-May 1941 u| http://archive.org/details/facultybulletinp00penn`

banukutlu:
Closes #301

### DIFF
--- a/lib/psulib_traject/macros.rb
+++ b/lib/psulib_traject/macros.rb
@@ -93,7 +93,7 @@ module PsulibTraject
       when 'full'
         fulltext_link_available?(field.indicator1, field.indicator2, url_label)
       when 'partial'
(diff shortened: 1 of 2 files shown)


nicktiberi (approved):
I must admit that I don't fully understand the relevance of all these rules, but this change makes sense to me and looks good! 👍

PR #380, opened 2021-12-15
Extract the endowment fund name so it can be used for display

nicktiberi:
Re: #57.

We extract the endowment fund name and code into a single Solr field for keyword search, but we only want to display fund name in the UI.


banukutlu (approved):
👌🏻

PR #390, opened 2022-03-16
Adds trim_punctuation to subject_other_display field

ajkiessl:
There were periods at the end of the subjects and I don't think we want them displaying that way in the Cat.  I found this 'trim_punctuation' parameter that seems to fix it.  I'm not sure where 'trim_punctuation' is defined though.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -249,6 +249,7 @@ PLATFORMS
   universal-java-10
   universal-java-11
   universal-java-14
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
lgtm 👍🏻  yup `trim_punctuation` is the right thing to use in this case, it is one of the Traject's built-in methods, thanks Alex.

PR #392, opened 2022-03-17
Adds equipment 949t mappings.  A fixture with one of these keys and a test.

### LINKED ISSUE #383: Add Equipment to formats
We're going to be adding equipment to the catalog. They will probably show up as "Book" under current logic. We should test the 949t to see whether it contains any of the following:

`LAPTOP or CARRELKEY or EBOOKREADR or EQUIP4HR or EQUIP24HR or EQUIP3DAY or EQUIP5DAY or EQUIP7DAY or EQUIP14DAY or EQUIP24FEE or EQIP7D5DF` - all these are equipment types.

ajkiessl:
I _believe_ this is all that is needed.  Looking through some of the logic, if a record has a 949t mapping to 'Equipment' (like EQUIP24HR), this will be given preference over the leader if it indicates that the record type is a 'Book'.  The test/fixture I added tests this.  It works with all the examples Ruth provided.

I can add more tests if we think we need to test the other equipment mappings (LAPTOP, CARRELKEY, etc.)

closes #383


banukutlu:
> I _believe_ this is all that is needed. Looking through some of the logic, if a record has a 949t mapping to 'Equipment' (like EQUIP24HR), this will be given preference over the leader if it indicates that the record type is a 'Book'. The test/fixture I added tests this. It works with all the examples Ruth provided.
> 
> I can add more tests if we think we need to test the other equipment mappings (LAPTOP, CARRELKEY, etc.)
> 
> closes #383

what you have looks good to me Alex, no need to test other equipment imo, thanks!

PR #394, opened 2022-03-29
Index 699a (Thesis Department) for faceting and advanced search

### LINKED ISSUE #368: Index Thesis Department for Advanced Search
Desired feature: A controlled drop-down in Advanced Search and a linked display field in record display for Penn State's Thesis Departments.

This information is in field 699a _but_ we can only guarantee that the data in 699a is **ours** if the 949t is also THESIS-B or THESIS-M or THESIS-D which is how we record our Penn State theses. (see bottom of issue for more)

We wouldn't need any translation table, it would just be generated based on the text of the field, e.g. History, Meteorology, Nuclear Engineering, Plant Sciences, etc. (we may want to test the kind of text we're getting and add thing like strip period off the end of the field if exists).

Logic:

```
if 949t == THESIS-B or THESIS-M or THESIS-D:
  index 699a as thesis department
```

Sample catkeys: `7829967,289554,28948319, 28948190`

(NOTE: one question out to a cataloger re: the depts that are lists... are they one dept as a list/and e.g. "Astronomy and Astrophycis" or is it two separate depts because then we need to check the data)

## WDLL

- [ ] thesis departments can be searched

See also: https://github.com/psu-libraries/psulib_blacklight/wiki/Thesis-Department

Reason why we need to test 949t: there are some electronic records with a 699a in them that's not connected to us. However we can't rule out all ONLINE records because our ETDAs are online.  However, the thesis records _all_ do have an item type of `THESIS-[BMD]` even if it's online, so we can use that for logic.

nicktiberi:
Closes #368.

### DIFF
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -576,6 +576,9 @@ to_field 'bound_with_struct' do |record, accumulator|
   end
 end
 
+# 699a Thesis Department
+to_field 'thesis_dept_facet', extract_marc('699a'), trim_punctuation, include_psu_theses_only
+
 # Place
 #
 # UP Library facet
--- a/lib/psulib_traject/macros.rb
+++ b/lib/psulib_traject/macros.rb
@@ -190,5 +190,26 @@ module PsulibTraject
         accumulator.compact!
       end
     end
+
+    def include_psu_theses_only
+      lambda do |record, accumulator|
+        accumulator.clear unless psu_thesis?(record)
+        accumulator.compact!
+      end
+    end
+
+    private
+
+      def psu_thesis?(record)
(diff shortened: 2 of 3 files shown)


config/traject.rb
```diff
@@ -576,6 +576,9 @@
   end
 end
 
+# 699a Thesis Department
+to_field 'thesis_dept_facet', extract_marc('699a'), trim_punctuation, include_psu_theses_only
```
banukutlu, line 580: @nicktiberi So, we are not checking if `psu_thesis?` for the facet field?

banukutlu (approved):
lgtm 👌🏻 😎

PR #399, opened 2022-04-06
Fix logic for ignoring dates out of valid range + add specs

### LINKED ISSUE #163: Remove date 9999 from consideration
The date 9999 is supposed to be used when things are open-ended. It's sometimes also used because of poor data choices.

We do not want it to carry any weight in:

1. date ranges (e.g. facet)
2. relevancy ranking

Example records?

nicktiberi:
Closes #163.

Previously, the logic to ignore dates `< 500` and `> current_year + 6` was not working properly, and was actually not ignoring any dates.

With this change, we are now actually ignoring dates `< 500` and `> current_year + 6`, which has the effect of ignoring all 9999 dates, but also ignoring other dates. We should confirm that this is acceptable to @ruthtillman before merging this in.


banukutlu (approved):
great catch ✔️  thanks @nicktiberi!

ajkiessl (approved):
Looks good 👍

PR #402, opened 2022-04-07
Use a frozen regex to compare PSU thesis codes to speed up indexing

nicktiberi:
The thought is that not allocating an array each time will reduce the performance bottleneck that indexing thesis codes introduced.


banukutlu (approved):
🚢 🤞🏻

whereismyjetpack:
After the current index completes, I'll run another one with this commit and report back!

PR #407, opened 2022-04-11
Libraries locations

ruthtillman:
Sample catkey: 17944044


lib/translation_maps/locations.properties
```diff
@@ -1,357 +1,9 @@
-ACAD-BLDG = Annexed Material
```
banukutlu, line 1: @ruthtillman why there are so many lines deleted?

banukutlu (changes requested):

PR #412, opened 2022-04-26
Adds logic to determine if 'Free to Read' in access_facet when multiple 506s

### LINKED ISSUE #302: Support multiple 506 for Free To Read
Currently if a record has one more than 506 , Free To Read check takes the first one, so the records with multiple 506s like the below examples does not resolve as Free To Read.

examples catkeys:

```
17532274
15647180
15707470
```

ajkiessl:
closes #302


banukutlu (approved):
👌🏻  looks great, thanks @ajkiessl

PR #414, opened 2022-04-26
Fix issues with some 856 URLs not being indexed for display

nicktiberi:
Closes https://github.com/psu-libraries/psulib_blacklight/issues/428

- allow any combination of upper/lowercase for `http(s)://` prefix
- prepend `http://` to URLs that start with `www.`


banukutlu (approved):
lgtm 👍🏻  thanks!

ajkiessl (approved):

PR #417, opened 2022-05-03
Modified regex for trimming call numbers to handle more permutations

### LINKED ISSUE #393: Additional things to trim from call numbers
Note: the problems might also involve the periods in them? In many cases, these have no space before the number!

* `panel.`: 13591327
* `svar.`: 1505296
* `sheet.`: 8165755, 280114 (the "t." seems to be getting caught, but not the rest?)
* `page.`: 13596094
* `e.`: 1278198
* `t.` (might be covered?): 1278198
* `roc.` : 199437, 74896, 127336, 1215261
* `n.s.`: 199386
* `cis.`: 74896, 1215261
* `dil`: 127680 (also varying "pt"), 1099910, 572217
* `Hft.`: 100152, 2223503
* `kn.`: 572042
* `knj.`: 118357, 187099
* `rel.`: 3047093
* `ses.`: 572217
* `année.`: 961969
* `Nr.`: 2223503
* `g.`: 114590
* `ano.` & `año.` - 58265, 343536
* `epoca` or `época` -- this tends to be formatted as "\d epoca" or "\d.epoca": 343536 & 110534
* `rik.`: 296153 (some have space, some don't)
* `leto`: 135451

ajkiessl:
I wonder if we could get the trimming to run the `#remove_by_regex` method in a loop until it no longer finds any more regex matches?  It looks like it's just trimming once, so if it matches to something at the end of the call number and misses something closer to the front of the call number, it doesn't do all the trimming it should.

closes #393

### DIFF
--- a/lib/psulib_traject/call_number.rb
+++ b/lib/psulib_traject/call_number.rb
@@ -91,7 +91,11 @@ module PsulibTraject
       def base_value
         case classification
         when 'LC', 'LCPER'
-          PsulibTraject::Processors::CallNumber::LC.new(value, serial: serial?).reduce
+          new_value = value
+          until PsulibTraject::Processors::CallNumber::LC.new(new_value, serial: serial?).reduce == new_value do
+            new_value = PsulibTraject::Processors::CallNumber::LC.new(new_value, serial: serial?).reduce
+          end
+          new_value
         when 'DEWEY'
           PsulibTraject::Processors::CallNumber::Dewey.new(value, serial: serial?).reduce
         else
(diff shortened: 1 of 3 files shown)


nicktiberi:
> I wonder if we could get the trimming to run the #remove_by_regex method in a loop until it no longer finds any more regex matches? It looks like it's just trimming once, so if it matches to something at the end of the call number and misses something closer to the front of the call number, it doesn't do all the trimming it should.

That makes a lot of sense to me and seems like it would give us more confidence that we are trimming everything we need to trim.

ajkiessl:
@banukutlu Ready to go!

banukutlu (approved):
thanks for working on this @ajkiessl  looks great!

PR #418, opened 2022-05-11
sort Dir.glob

whereismyjetpack:
On some systems, Dir.glob does not return a sorted list, this could cause us to process incrementals out of order when we replay them. 

Calls .sort on Dir.glob calls to return a sorted list.


ajkiessl (approved):

banukutlu (approved):
👍🏻

PR #437, opened 2022-11-16
Parse more genres and separate multiple internal genres

### LINKED ISSUE #185: Add 655 rbgenr to indexed genres
When we index genres, we test to see if they're in one of a few vocabularies.

Processing test is here: https://github.com/psu-libraries/psulib_traject/blob/main/lib/psulib_traject/macros.rb#L14-L29

vocabulary list is here: https://github.com/psu-libraries/psulib_traject/blob/main/lib/psulib_traject/macros.rb#L19

We want to add several vocabularies to support rare books. These are:

```
rbgenr rbbin rbprov rbpub rbpri aat rbmscv 
```

We do not currently have any rbmscv, but it's currently being worked on as a way to combine all the vocabs that begin with "rb," (if you were 👀 the number of rare book terms in the list, so are the folks responsible for it, but they won't be done that project for a while) so this will save us a future ticket in a year or two when those start coming in.

That part should be easy, but it will then create some multi-part genres, e.g. https://catalog.libraries.psu.edu/catalog/159973 which has: |a Periodicals |y 20th century. 

We will want to be sure they display ok, BUT **we do not care about treating them like we do multi-part subjects, just treat them as a single string**. It's just a matter of being sure they look ok/click through to search ok!

Sample records:

-7531373
-9365798
-4017726
-516392
-516352
-34870166
-34870198
-34874802
-35140017
-35140021
-35140023
-36212453
-36478257
-197501
-197524

## WDDLL 

* [x] Genres from these vocabs display in records with link to search
* [x] Clicking on one of these takes you to a controlled-value search for that term/phrase in genres
* [x] Genres from these vocabs appear in sidebar facets.

ajkiessl:
Also, adds more subfields to be parsed for the genre facet.  The same subfields the main genre field uses.

closes #185

### DIFF
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -265,7 +265,7 @@ to_field 'subject_browse_facet', process_subject_browse_facet(
 to_field 'genre_tsim', extract_marc('650|*0|v:655|*0|abcvxyz:655|*7|abcvxyz')
 
 ## Genre facet (sidebar)
-to_field 'genre_facet', process_genre('650|*0|v:655|*0|abcvxyz:655|*7|ay')
+to_field 'genre_facet', process_genre('650|*0|v:655|*0|abcvxyz:655|*7|abcvxyz')
 
 ## Genre display
 to_field 'genre_display_ssm', process_genre('655|*0|abcvxyz:655|*7|abcvxyz')
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -265,7 +265,7 @@ to_field 'subject_browse_facet', process_subject_browse_facet(
 to_field 'genre_tsim', extract_marc('650|*0|v:655|*0|abcvxyz:655|*7|abcvxyz')
 
 ## Genre facet (sidebar)
-to_field 'genre_facet', process_genre('650|*0|v:655|*0|a:655|*7|a')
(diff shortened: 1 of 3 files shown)


config/traject.rb
```diff
@@ -265,13 +265,13 @@
...
 ## Genre facet (sidebar)
-to_field 'genre_facet', process_genre('650|*0|v:655|*0|a:655|*7|a')
+to_field 'genre_facet', process_genre('650|*0|v:655|*0|abcvxyz:655|*7|abcvxyz')
 
 ## Genre display
 to_field 'genre_display_ssm', process_genre('655|*0|abcvxyz:655|*7|abcvxyz')
 
 ## For genre links
-to_field 'genre_full_facet', extract_marc('650|*0|v:655|*0|abcvxyz:655|*7|abcvxyz'), trim_punctuation
+to_field 'genre_full_facet', process_genre('650|*0|v:655|*0|abcvxyz:655|*7|abcvxyz')
```
ajkiessl, line 274: I had to make this change to get the search to work when clicking the link.  Idk if there was a benefit to just using 'extract_marc' over the 'process_genre' method.

banukutlu: it makes sense to make this change 👍🏻

PR #438, opened 2022-11-16
Split 653a on semicolon

### LINKED ISSUE #416: Break up multi-value 653s on semicolons

ajkiessl:
Stripping leading and trailing whitespace from 653a's and splitting 653a into separate items on ';'s.

closes #416


banukutlu (approved):
👍🏻

PR #439, opened 2022-11-17
Adds 773 extract with trim_punctuation

### LINKED ISSUE #433: Index MARC 773 for display as "part of"
Special Collections would like to add the MARC 773 (with first indicator 0) to the index for display in records. We'll be calling this "part of" because that's probably the best name for it.

Traject code would be:

```
773|0*|abdghiklmnopqrstuwxyz34
```

The field may repeat and we would display separate entries on successive lines. For display, see: 

## WDDLL

- [ ] the 773s are indexed for display as text in a record.

## Sample Catkeys

```
4441570
5017247
32863792
2008160
1983887
1894771
1852785
1737126
```

ajkiessl:
closes #433 

blacklight: https://github.com/psu-libraries/psulib_blacklight/pull/1082

### DIFF
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -575,6 +575,9 @@ end
 # 699a Thesis Department
 to_field 'thesis_dept_facet', extract_marc('699a'), trim_punctuation, include_psu_theses_only
 
+# 773 - "Part Of"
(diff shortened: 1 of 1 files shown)


banukutlu (approved):
👍🏻

PR #441, opened 2022-11-21
Removed all subfields except t subfields in serials traject fields

### LINKED ISSUE #421: Continues / continued by linking without ISSN
Right now, the continues/continued by is including the ISSN _if it exists_, in the title search, e.g. https://catalog.libraries.psu.edu/?search_field=title&q=PennState+agriculture+0889-5929

Sometimes it's just a title. This record has a good example of how it could be either way: https://catalog.libraries.psu.edu/catalog/313610

Because we are specifically searching title, the ISSN doesn't work. The title is in the `subfield |t`. Fixing this might just look like only indexing and displaying the title vs title and ISSN. Or it might look like adding a separate field. Either is fine.

## WDDLL

- [ ] Title search link only contains title so it's working well. Change the marc mapping for all serial fields so that it only maps subfield `t` instead of `iabdghkmnopqrstuxyz3` https://github.com/psu-libraries/psulib_traject/blob/f600c4288041dead740693f20e9d907b92b385a3/config/traject.rb#L601-L614

ajkiessl:
Removes all subfields from serials marc extraction except 't' (title).

closes #421

### DIFF
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -599,19 +599,19 @@ to_field 'dedication_ssim', extract_marc('995abcd3n')
 # Serials fields
 #
 # Preceding and Succeeding Entries display
-to_field 'serials_continues_display_ssim', extract_marc('780|00|iabdghkmnopqrstuxyz3:780|02|iabdghkmnopqrstuxyz3')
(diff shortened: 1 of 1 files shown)


banukutlu (approved):
👍🏻

PR #446, opened 2022-12-08
use jdk17

whereismyjetpack:
updates JDK for traject. We should be able to cut a release of traject, and do a full index, we'll test indexing in to the -green deployemnt first. we've indexed into qa multiple times succesfully after these tweaks, and tweaks to the catalog config


banukutlu (approved):
👍🏻  thanks!

PR #447, opened 2022-12-09
Track hourlies by collection, not alias 

whereismyjetpack:
- figures out what the collection name is if using an alias
- uses that collection name as a key in redis to track hourlies

This allows us to process hourlies _before_ flipping the collection on a full re-index.


banukutlu:
> * figures out what the collection name is if using an alias
> * uses that collection name as a key in redis to track hourlies
> 
> This allows us to process hourlies _before_ flipping the collection on a full re-index.

oh this is a great improvement! thanks Dann!

banukutlu (approved):
👌🏻 :shipit:

PR #458, opened 2023-03-29
Adds iiif_manifest_ssim field

### LINKED ISSUE #456: Add IIIF Manifest URL to its own field
In order to integrate a IIIF viewer, we'll need to add IIIF manifest URLs to our records and index them. This field will **not** be displayed. Only the 856u value needs to be in the index.

These fields can be can be identified through the following logic:

IF:

    `856|*1` AND subfield y == `IIIF manifest`

THEN:

    index the URL (856u)

We do not need to index the labels, etc.

So essentially they'll match what we're calling "partial" and then we need to split off the subset which have a subfield y of `IIIF manifest` (this is the specific case we'll be using, but we could downcase it to be safe). While it's unlikely that a single record will have more than one IIIF manifest, this is a theoretical possibility so we'll want to treat it as a possibly repeating field.

Sample catkeys:

```
2025781
7280921
7659409
11939133
35139815
1267921
```

## WDDLL

- [ ] On these records, the index contains a IIIF-specific field with the URL's value.

banukutlu:
Closes #456


.ruby-version
```diff
@@ -1 +1 @@
-jruby-9.2.11.1
+2.6.6
```
whereismyjetpack, line 1: Do we want ruby version to be jruby?

whereismyjetpack (approved):

.ruby-version
banukutlu: yessss! good catch :) thanks! will fix it

whereismyjetpack (approved):

whereismyjetpack:
I'm going to re-index qa with this code.

PR #472, opened 2023-05-15
Extract 830s for series title display

### LINKED ISSUE #313: Update Series Title to include 830, some conditional logic
This has been a weird one, and one I should've gotten to back when I made the ticket, but I didn't and here we are. Right now it's at: https://github.com/psu-libraries/psulib_traject/blob/main/config/traject.rb#L162-L164 We're going to remove 440s entirely as they've been phased out. Any that are leftover are errors and should be ignored. 

Can we update the indexing logic to be:

For search:

abbreviated: `440a:490a:830a` 

fuller: `440av:490anpv:830anpv` 

For display:

- IF there is an 830:
  - THEN index `830adfghklmnoprstvwxy3` into series title display
       - AND if there is also a 490
         - IF the 490's first indicator is blank or 0 (so `0*` or ` *` -- I think that space is correct for blank?) then index `490avlx` into series title display as well.
         - ELSE if the 490's first indicator is 1, then ignore it.
- ELSE if there is no 830 but any kind of 490:
  - THEN index `490avlx` into series title display, no matter what the indicators
- AND if there is any 44:
  - THEN index it to series title display as `440anpvx`

Basically, the existence of an 830 means there's a more formal version of a 490|1*| but if there is a 490|0*| then it's unrelated to the 830 and represents a second series. We'll display just one. 440s should not exist at all, but they do, so if they're there then there won't be a 490 and almost certainly won't be an 830 and we still need to represent the series. While the 830 is the formal version, someone might do a phase search on what's in the 490, which is why we're retaining it for search.

And we're not displaying those 490|1*| because they should look very similar to but slightly different from the 830s and it would be a weird experience for someone to see like

Sample records:

- 2702238
- 180
- 509
- 629307
- 1152735

(yes some VERY old catkeys, I did a query and ordered results by catkey)

ajkiessl:
@ruthtillman Could you double-check this logic?  

when only an 830 field is present:
  stores the 830 data in the series_title_display_ssm

when an 830 field and 490 field are present but 490 does not get appended (because indicator1 = '1'):
  stores the 830 data in the series_title_display_ssm

when an 830 field and 490 field are present (490 _is_ appended because indicator1 is either ' ' or '0'):
  stores the 830 and 490 data in the series_title_display_ssm

when an 830 field and 440 field are present:
  stores the 440 data in the series_title_display_ssm

when only a 490 field is present:
  stores the 490 data in the series_title_display_ssm

when a 490 field and 440 field are present:
  stores the 440 data in the series_title_display_ssm

when only a 440 field is present:
  stores the 440 data in the series_title_display_ssm

closes #313


ajkiessl:
Also, @ruthtillman idk if you would know this, but does the "abbreviated" version of the series title map to the "strict" version we store in Solr?  That's how I set it up here:

https://github.com/psu-libraries/psulib_traject/pull/472/files#diff-568723dfd47f0b65f4ac3c98258e73761ec600084983f4f3dc3b8570d605441eR163-R164

Smullz622 (approved):
LGTM!

ruthtillman:
@ajkiessl Looked at the data and updated the logic in the issue logic. There are one or two cases where this could lead to messiness but those are cases where we'll need to fix the data.

PR #475, opened 2023-05-19
right to left language formatting

### LINKED ISSUE #176: Periods at the wrong end of right-to-left transliterated languages
e.g. Arabic. Affected fields include edition and publication. The field is displaying correctly right-to-left, but the period is on the right side, as in English. Example catkeys:

- 22342875
- 12145197

Smullz622:
Fixes #176 

New processor moves period to the left side for vernacular publication fields in right to left languages

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -251,6 +251,7 @@ PLATFORMS
   universal-java-14
   universal-java-15
   universal-java-17
+  universal-java-19
   x86_64-darwin-20
   x86_64-linux
 
@@ -281,4 +282,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   2.2.20
+   2.2.33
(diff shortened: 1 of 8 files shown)


Gemfile.lock
```diff
@@ -281,4 +282,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   2.2.20
+   2.2.33
```
ajkiessl, line 285: It's up to you if you want to change this and it may not be an issue, but to be safe we might want to keep the bundler at 2.2.20.  If it deploys and there are no issues, then I think it's fine.

ajkiessl (approved):
Other than my one comment below, this LGTM.

Gemfile.lock
Smullz622: Oh weird, I missed that. It must have happened when I ran bundle install to get tract up to date. I'll change it back.

ajkiessl (approved):

PR #478, opened 2023-06-07
upgrading to JRuby 9.4.2.0

Smullz622:


whereismyjetpack (dismissed):

ajkiessl:
I noticed there's a change in the travis.yml.  Can we just delete the travis.yml since it is not being used anymore?  Alsp, should we consider bumping the ruby version in the circleci.yml? I haven't been able to fully dig into the circleci setup, but I'm not sure if the defined ruby version in there is the version being used by the tests/linting.

ajkiessl (changes requested):
Yeah, looking at it some more, I think this should be bumped to something in the 3.1.x range:

https://github.com/psu-libraries/psulib_traject/blob/main/.circleci/config.yml#L41

whereismyjetpack:
travis can go. I'm not sure why it's even there!

whereismyjetpack (approved):

ajkiessl (approved):
LGTM

PR #488, opened 2023-11-03
Sidekiq update

Smullz622:
Bumping sidekiq required sidekiq-scheduler to be upgraded as well.

The hourly indexer spec was failing. It seemed like it was only removing keys that had `hr` but none of them did, so it wasn't actually removing anything.


ajkiessl (approved):
Looks good 👍

PR #493, opened 2023-11-30
Combine biographical sketch and administrative history fields into one

### LINKED ISSUE #490: Change indexing for 545
Right now we're indexing the 545 two different ways based on indicators. I've gotten feedback that the no-indicator case is showing a weird label on a number of our archival records (where this will be used most). Just switching it the other way won't work, so the next best choice is to make it one field with a label saying it could be either.

So instead of:

https://github.com/psu-libraries/psulib_traject/blob/main/config/traject.rb#L519-L521

we'll just index:

```
to_field 'bio_admin_sketch_note_ssm', extract_marc('545abu')
```

ajkiessl:
closes #490


Smullz622 (approved):
lgtm

PR #495, opened 2024-01-02
laser disc logic removed

### LINKED ISSUE #494: A subset of records about Caving are showing up as Laser discs...
Most of our "these are not laser discs" problems are created by records that were originally created for laser discs that then got reused for DVDs and Blu Rays. These make sense and the problem has been handed over to the cataloging department for fixing.

However, we were getting a subset which were really throwing me. I finally figured out it was the 538, which was doing a case-insensitive search for "CAV" among other things, which makes sense but is catching other things. Like caves!

Given how rare actual laser discs are, it's ok to rely on the other two tests in the 007 and the 949. Let's get rid of this line entirely. The other tests from this section are much less likely to appear in other words and we're more likely to have them, so we should keep them (and some of them are helping records that might otherwise think they were only Laser discs show up as both Laser disc and Blu-ray, which means at least Blu-ray searchers can find them).

Line to remove: https://github.com/psu-libraries/psulib_traject/blob/main/lib/psulib_traject/processors/media_type.rb#L73

Sample records:

- 2696747
- 2697440
- 2871565
- 3171524
- 3500525 (also this one where another "*cav*" is in the 538 URL.

Smullz622:
Fixes #494 

Logic was falsely identifying items with 'caving' as laser disc. Since there are other fields that can identify laser discs, we're just removing this logic.

### DIFF
--- a/lib/psulib_traject/processors/media_type.rb
+++ b/lib/psulib_traject/processors/media_type.rb
@@ -70,7 +70,6 @@ module PsulibTraject::Processors
           results << 'Blu-ray' if field_538a&.match?(Regexp.union(/Bluray/i, /Blu-ray/i, /Blu ray/i))
           results << 'Videocassette (VHS)' if /VHS/i.match?(field_538a)
           results << 'DVD' if /DVD/i.match?(field_538a)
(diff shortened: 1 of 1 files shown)


ajkiessl (approved):
🚀

PR #504, opened 2024-04-25
Adds 023 MARC field to Solr fields

jlandiseigsti:
023a and 023z are now added to the Solr field 'issn_sim'. Additionally, 023a is added to the display field 'grouping_issn_ssm'.


ajkiessl (approved):
lgtm

PR #506, opened 2024-05-09
503 - Grabs 857 field if it starts with ark

### LINKED ISSUE #503: Index 857u for use in Aeon requests
We will need to index 857u subfield for our LIBSAFE materials, if they exist, for use in Aeon requests. These will only be used to power requests, not displayed as they are not public links (this system allows folks to determine if the person can be sent a copy of scan we have already).

Note: we already have a SubLocation field, so we might want to index this to that and/or append it to that? https://github.com/psu-libraries/psulib_traject/blob/main/config/traject.rb#L626-L629

In order to ensure we only index our data, we will only want to index if the field _begins with_ our ARK prefix:

```
ark:/42409
```

Sample record is only on our QA system at this point: 17289297

can be seen in MARC view here: https://catalog.stage.k8s.libraries.psu.edu/catalog/17289297

jlandiseigsti:
closes #503 
If 857 starts with our ark prefix, adds it to the sublocation ssm.

### DIFF
--- a/config/traject.rb
+++ b/config/traject.rb
@@ -629,7 +629,15 @@ to_field 'serials_changed_back_to_display_ssim', extract_marc('785|08|t')
 # 799a - Sublocation
 # From our catalog experts: "The data in this field corresponds to "collections" within Special Collections, and
 # the 799 data lets their staff know which shelf (or range of shelves) to check for the call number in question.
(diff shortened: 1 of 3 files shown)


Smullz622 (approved):
looks good, thanks!

ajkiessl (approved):

PR #507, opened 2024-05-10
Index SUDOC values

### LINKED ISSUE #406: Identify Base SUDOC call numbers
Part of #285

In SUDOC, we can expect letters and numbers separated by one of: `:-./` and possibly spaces. However, there MAY not be a space before the parts information, e.g.

- D 101.47:v.14
- C 3.2:H 62/8/v.2

These would be after the last separator of some kind, however.

We should expect the most common volume, issue, and number, v. iss. vol. num. etc words and we only expect English on these.

Also, if we are removing things like "KIT" from call numbers, we will want to bring that over here, e.g..

- HE 20.427:C 76/2/KIT. 
- HE 20.427:H 33 
- HE 20.427:H 75/PACK. 
- HE 20.427:IL 6/KIT 
- HE 20.427:IN 7 

## WDLL
- [ ]  Generate base call numbers for SUDOC call numbers

ajkiessl:
fixes #406 

Now indexing SUDOC values.  The next step of this would be to start identifying what kind of processing needs done on these.  We may want to set up a preview branch for this and have it do a full index.  Then output a large subset of these for Ruth to look at.

### DIFF
--- a/lib/psulib_traject/call_number.rb
+++ b/lib/psulib_traject/call_number.rb
@@ -27,11 +27,13 @@ module PsulibTraject
     end
 
     def normalized_shelfkey
+      return shelf_key.call_number if sudoc?
+
       shelf_key.normalized
     end
 
     def not_browsable?
-      return true unless lc? || dewey?
+      return true unless lc? || dewey? || sudoc?
 
       normalized_shelfkey.nil?
     end
@@ -100,6 +102,10 @@ module PsulibTraject
 
       def dewey?
         classification == 'DEWEY'
+      end
+
+      def sudoc?
+        classification == 'SUDOC'
       end
 
       def classification_to_field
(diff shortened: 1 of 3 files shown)


lib/psulib_traject/call_number.rb
```diff
@@ -102,6 +105,10 @@ def dewey?
         classification == 'DEWEY'
       end
 
+      def sudoc?
+        classification == 'SUDOC'
```
jlandiseigsti, line 109: I like these shortcuts.

jlandiseigsti: Is there any possibility the returned case would ever be different (eg 'Sudoc')? I assume not, but just wanted to check.

jlandiseigsti (approved):

lib/psulib_traject/call_number.rb
ajkiessl: Hmm, good question.  I would say no since the others don't do any case normalization.  It's probably some standard for marc that they are always in uppercase, but I wouldn't be against adding case normalization to all of these to be safe.

Smullz622 (approved):

PR #515, opened 2024-12-17
update 2nd floor Pattee labels

jackflaps:


ruthtillman:
Closing without merging because we caught up ahead.

PR #523, opened 2025-02-28
Supress 700 and 710 fields if owner is found

### LINKED ISSUE #518: Suppress a subset of 700 and 710 fields from facet/display based on |e data
We need to suppress a subset of 700 and 710 (but not 711 I'm so sorry) based on testing for whether the |e contains "owner".

We want to remove them from:
* `author_addl_tsim`
* `all_authors_facet`
* `addl_author_display_ssm`

(I think that's everywhere we're using 700 and 710, but might be worth checking)

We are not looking at 1xx fields or 711. Ideally the solution would allow us to test for additional terms in future (so, array).

### Sample Records

* 39991821
* 38979793

## Longer explanation

700 and and 710 are for personal or corporate persons who are somehow related to the work but not the primary creator. 99% of the time these are used for additional authors or creators. `|e` is the "relator" subfield and normally describes in greater specificity how a person contributed to the creation of the work -- e.g. a movie's costume designer or a book's illustrator or even something niche like the "land surveyor" for a map. However, there are a very small subset of relator fields that may describe the way a person "relates" to a work that isn't actually about creating it.

The field is also used in 100 and 110 fields for primary people related to a work and can only occur once per record. So even if we had a person who was related in a non-creator way, if they're in the 1xx fields, they're the best/only value someone could find for that record and we'll keep them.

In 111 and 711, the subfield |e means something else 🙃 so we aren't looking at 711.

Up until now, this has been such a small part of our catalog data that it didn't matter. But with the cataloging of the Lee Bennett Hopkins project, we have run into an issue where we want to add the field to explain that he was the "former owner" of these books. This will impact several thousand books. Meanwhile, we also want to be able to highlight the books he actually wrote or edited, which our librarians estimate as being about 100-200. Indexing the former owner fields is unfortunately causing him to show up as one of the creators + we're getting all the books in this faceted search: https://catalog.libraries.psu.edu/?f[all_authors_facet][]=Hopkins%2C+Lee+Bennett

This is one of those cases where we haven't been impacted by a really really edge case until suddenly we are. Even though it only impacts a few thousand records, they're important to the PA Center for the Book and Special Collections both. I chose "owner" because that would get any of the other owner/non-creator variations including "copyright owner" (which we do have but I haven't looked into the record). I did a review of unique field values for this subfield and I didn't find anything else I think we need to include at this time, but I specified that it should be an array so we can easily update if needed. (I did find a million typos)

memsdm05:
Fixes #518

### DIFF
--- a/lib/psulib_traject/macros.rb
+++ b/lib/psulib_traject/macros.rb
@@ -17,7 +17,7 @@ module PsulibTraject
 
         Traject::MarcExtractor.cached(fields).collect_matching_lines(record) do |field, spec, extractor|
           next if field.subfields.any? { |subfield| subfield.code == 'e' && subfield.value.include?('owner') }
-          
+
           values = extractor.collect_subfields(field, spec)
           accumulator.concat(values) if values && !values.empty?
         end
--- a/lib/psulib_traject/macros.rb
+++ b/lib/psulib_traject/macros.rb
@@ -8,6 +8,22 @@ MAX_YEAR = Time.new.year + 6
 
 module PsulibTraject
   module Macros
+    # Extracts facets without owner field
+    #
+    # Checks to see if owner is in e subfield, if so it skips
+    def extract_marc_without_owner(fields)
+      lambda do |record, accumulator|
+        return nil unless record.is_a? MARC::Record
+
+        Traject::MarcExtractor.cached(fields).collect_matching_lines(record) do |field, spec, extractor|
+          next if field.subfields.any? { |subfield| subfield.code == 'e' && subfield.value.include?('owner') }
+          
+          values = extractor.collect_subfields(field, spec)
+          accumulator.concat(values) if values && !values.empty?
+        end
+      end
+    end
+
     # For genre facet and display
     #
     # limit to subfield $2 vocabularies for 655|*7 genres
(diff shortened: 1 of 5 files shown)


lib/psulib_traject/macros.rb
```diff
@@ -8,6 +8,22 @@
...
+    # Checks to see if owner is in e subfield, if so it skips
+    def extract_marc_without_owner(fields)
+      lambda do |record, accumulator|
+        return nil unless record.is_a? MARC::Record
+
+        Traject::MarcExtractor.cached(fields).collect_matching_lines(record) do |field, spec, extractor|
+          next if field.subfields.any? { |subfield| subfield.code == 'e' && subfield.value.include?('owner') }
+
+          values = extractor.collect_subfields(field, spec)
+          accumulator.concat(values) if values && !values.empty?
```
Smullz622, line 22: could this be consolidated to `if values.present?`

Smullz622 (approved):
Small note you can take or leave, otherwise looks good. Great work!

lib/psulib_traject/macros.rb
memsdm05: I changed it to values.present? but now it errors out. If it ain't broke...

PR #524, opened 2025-03-07
Handle colon when sorting call numbers

memsdm05:


memsdm05:
Moving implementation to blacklight

PR #525, opened 2025-04-02
Sort LCN with colon in cutter

### LINKED ISSUE #405: Call number browse - sorting colon in call number
# Bug Report

## Description
The only feedback we've gotten on call number browse sorting order was related to colons in call numbers. These are most likely to occur in maps. 

## To Reproduce
You can see the colon-type call numbers starting here:

https://catalog.libraries.psu.edu/browse/call_numbers?classification=lc&nearby=G3824.S8

and then the normal .S8 call numbers that they should be following start here:

https://catalog.libraries.psu.edu/browse/call_numbers?classification=lc&length=50&starting=G..3824.S8.0002.P4.E67--1996C000400210001MPMCU

## Expected behavior

- [ ] Call numbers such as `G3824.S8:2A7E63 2011.P8` should sort after the rest of that call numbers such as `G3824.S8 1942.S8` (and similar). Paige A., maps cataloger, says this `:` should essentially sort after `Z`.

memsdm05:
Fixes #405


Smullz622 (approved):
LGTM! Thanks for adding those rake tasks!

